PR #36, opened 2015-06-03
A license.  Tada!

werebus:
It'd probably be a good idea :tm: to include a mention of the license in the README, but the README also needs a bit of fleshing out.


dfaulken:
Thanks @werebus

PR #38, opened 2015-06-06
More precise show_message testing

dfaulken:
@Anbranin this will affect how you implement your coming PR.


dfaulken:
The biggest change here is the [spec_helper method](https://github.com/umts/jobapps/pull/38/files#diff-93830fa29d616f7c87903d08b5b1b29aR38).

PR #41, opened 2015-06-09
Mailer view

Anbranin:
Made a new mailer view for interview rescheduling, and also wrote tests for mailer.


dfaulken:
I notice this reduces our coverage percentage; I'll try and root out where that's coming from and cover it before merging this.

PR #47, opened 2015-06-15
database.yml is a little platform-specific

werebus:
And some people might need passwords in there.


werebus:
Well, except for that, I guess.  Must be some other way.

PR #48, opened 2015-06-19
Adding seeds

dfaulken:


dfaulken:
I accidentally threw some styling in here instead of in another PR, but well. Here's an example of how it looks:

screen shot 2015-06-19 at 4 07 42 pm

PR #54, opened 2015-07-07
No user

dfaulken:
Staging this in a PR so that I can apply the same changes to ft-forms using a relatively similar tactic.


dfaulken:
Above done; merging

PR #57, opened 2015-07-15
View tests

dfaulken:
Opening on behalf of @Anbranin

### DIFF
--- a/spec/views/dashboard/application_templates_spec.rb
+++ b/spec/views/dashboard/application_templates_spec.rb
@@ -1,0 +1,50 @@
+require 'rails_helper'
+include RSpecHtmlMatchers
+
+describe 'dashboard/_application_templates.haml' do
+  before :each do
+    @department = create :department
+    assign :departments, Array(@department)
+    @position = create :position, department: @department
+    assign :positions, Array(@position)
+  end
+  context 'there are application templates present' do
+    before :each do
+      @application_template = create :application_template, position: @position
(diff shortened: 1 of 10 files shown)


spec/views/dashboard/application_templates_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+    it 'contains a link to edit the application template' do
+      render
+      action_path = edit_application_template_path @application_template
+      expect(rendered).to have_tag 'a', with: { href: action_path }
+    end
+  end
+  context 'there are no application templates for the positions listed' do
+    before :each do
+      # seems dumb to make an application template where there are none
```
dfaulken: What would better is to reference the part of the page that has to do with that position and make sure it doesn't have links.

Or better yet, since you can be sure it's the only position that'll be shown on that view:

```ruby
it 'has no links' do # since there are no positions with application templates
  render
  expect(rendered).not_to have_tag 'a'
end
```

dfaulken:
Very good work. I'll actually take care of the thing I mentioned above and merge this in. :+1:

spec/views/dashboard/application_templates_spec.rb
dfaulken: Actually never mind, since there is a link there. I'm content with just having there be the test for the presence of the link that *should* be there, and not worrying about what *shouldn't*.

PR #61, opened 2015-07-22
made application_templates show have default values for personal info…

Anbranin:
…rmation when current user exists. wrote spec tests for that.


dfaulken:
Lovely

PR #65, opened 2015-08-14
Draft Renaming and Testing

### LINKED ISSUE #58: Template drafts
It occurred to me yesterday that I never got around to adapting ft-forms' FormDraft system for use here. We should talk to Don about how high-priority configuring that application is for him - if it's not, then I think it'll make a good next project for @Anbranin. If it is then I'll take care of it ASAP.

Anbranin:
Created a draft model for applications much like in ft-forms. Wrote corresponding views, and removed any logic associated with editing application templates directly. Removed the "move question" logic from the model to the controller. Changed the class name from "application template drafts" to "application drafts." Finally, refactored code for readability, line length, and passing tests.

EDIT (@dfaulken): This PR closes #58.


dfaulken:
Would you mind updating the name and description of this PR?

Anbranin:
What kind of name would be better?

Anbranin:
Also, at this point, does the name matter? I can delete the branch after
it's merged

dfaulken:
You *should* delete the branch after it's merged. A good PR name and description are good practice, particularly a name that isn't an abject lie.

app/controllers/application_drafts_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+    redirect_to edit_draft_path
+  end
+
+  def remove_question
+    question_number = params.require(:number).to_i
+    @draft.remove_question question_number
+    redirect_to edit_draft_path
+  end
+
+  def save_question
```
dfaulken: Don't need this, right?

app/models/question.rb
```diff
@@ -18,8 +20,12 @@ class Question < ActiveRecord::Base
   # No questions in one application template with the same number
   validates :number, uniqueness: { scope: :application_template,
                                    allow_blank: true }
+  validates :number,
+            uniqueness: { scope: :application_draft,
+                          if: -> { application_draft.present? } }
 
   default_scope { order :number }
+  scope :not_new, -> { where.not id: nil }
```
dfaulken: I'd be open to suggestions of a better AR scope name here, I don't like `not_new`.

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 99.55
+    "covered_percent": 94.54
```
dfaulken: That's a dip, what isn't covered?

spec/models/application_template_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+  end
+  describe 'draft_belonging_to' do
+    before :each do
+      @user = create :user
+      other_user = create :user
+      @application_template = create :application_template
+      @draft = create :application_draft,
+                      application_template: @application_template,
+                      user: @user
+      # other draft - why do we need this?
```
dfaulken: Good question - so that we can be sure the method is returning what we expect it to, and not what we don't expect it to. The only way we can do this is by making data we expect it not to return. Say, for example, if the method just returned `ApplicationDraft.all`.

spec/models/application_template_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  describe 'draft_belonging_to?' do
+    before :each do
+      @application_template = create :application_template
+      @user = create :user
+    end
+    let :call do
+      @application_template.draft_belonging_to? @user
+    end
+    it 'returns false if a draft exists for the user in question' do
```
dfaulken: Wait, isn't that backwards? Doesn't it return `true` if there *is* one?

spec/views/application_drafts/edit_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    render
+    action_path = edit_draft_path @draft
+    expect(rendered).to have_form action_path, :post do
+      with_tag 'form', with: { class: 'edit_application_draft' }
+    end
+  end
+  it 'has inputs for each question field' do
+    render
+    expect(rendered).to have_tag 'tr' do
+      base_tag_name = 'application_draft[questions_attributes][2]'
```
dfaulken: Why 2?

dfaulken:
Very solid work, just a few items. :+1:

dfaulken:
Builds pass; deploying.

PR #72, opened 2015-08-19
App record show edit

Anbranin:
screen shot 2015-08-19 at 10 39 49 am

Closes #72


app/assets/stylesheets/application_records/show.scss
```diff
@@ -0,0 +1,28 @@
+/*
+desired behavior: 
+
+Concatenation
```
dfaulken: We can probably remove this.

PR #85, opened 2015-08-27
Default interview location

Anbranin:


coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 94.81
+    "covered_percent": 94.79
```
dfaulken: It'd be good to take a look at this and see why coverage percentage is declining.

dfaulken: I looked, and I'm actually not sure why. None of the < 100% covered files are touched in this PR.

PR #86, opened 2015-08-27
App rec show edit

### LINKED ISSUE #82: In ApplicationRecord#Show, questions and answers don't line up
Don pointed this out to me. When someone skips a question and leaves the response blank, the next response fills its space so it looks like it's lined up with that question they skipped. If enough questions are skipped, it is nigh-unreadable. I told Don what was happening and he says he will just look at the application records very closely until we fix it.

Anbranin:
Closes #82


dfaulken:
I see model changes but no specs... any chance you accidentally took some of the wrong changes with you?

PR #88, opened 2015-09-04
App drafts controller testing

### LINKED ISSUE #81: Test ApplicationDraftsController
Currently at 31.43%. It'd be nice to bring app-wide test coverage up above 99.5%.

I also submitted a support request to codeclimate about GH integration for coverage percentage changes.

Anbranin:
closes #81

### DIFF
--- a/spec/controllers/application_drafts_controller_spec.rb
+++ b/spec/controllers/application_drafts_controller_spec.rb
@@ -200,10 +200,13 @@ describe ApplicationDraftsController do
 
   describe 'POST #update' do
     before :each do
+      @commit = 'Save changes and continue editing'
       @draft = create :application_draft
+      @new_user = create :user
+      @draft_changes = { user_id: @new_user.id }
     end
     let :submit do
-      post :update, id: @draft
+      post :update, id: @draft, draft: @draft_changes, commit: @commit
     end
     context 'student' do
       before :each do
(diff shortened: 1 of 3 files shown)


dfaulken:
What's the coverage percentage on that particular file? 98.28% is still a little low.

werebus:
Just took a look: 82.86%  Looks like the `ApplicationDraftsController#update` action isn't being covered yet.

screenshot_2015-09-04_15-41-22

dfaulken:
Thanks @werebus. @Anbranin, I'll look over the tests that are here, if you wouldn't mind testing that last method and then pushing that up.

spec/controllers/application_drafts_controller_spec.rb
```diff
@@ -0,0 +1,271 @@
...
+        submit
+        expect(response).to have_http_status :unauthorized
+      end
+    end
+    context 'staff' do
+      before :each do
+        when_current_user_is :staff
+      end
+      it 'creates a draft for the correct application template' do
+        expect { submit }
```
dfaulken: Hmm, I'd also like to mock `ApplicationTemplate#create_draft` - I like that the student scenario tests for its absence, so we should check for its presence here.

spec/controllers/application_drafts_controller_spec.rb
```diff
@@ -0,0 +1,271 @@
...
+      before :each do
+        when_current_user_is :staff
+      end
+      it 'assigns the correct draft to the draft instance variable' do
+        submit
+        expect(assigns.fetch :draft).to eql @draft
+      end
+      it 'calls #remove_question on the draft with the number referenced' do
+        expect_any_instance_of(ApplicationDraft)
+          .to receive(:remove_question)
```
dfaulken: Here's a good example of what I meant above.

spec/controllers/application_drafts_controller_spec.rb
```diff
@@ -0,0 +1,271 @@
...
+        expect(response).to have_http_status :unauthorized
+      end
+    end
+    context 'staff' do
+      before :each do
+        when_current_user_is :staff
+      end
+      it 'updates the draft with the changes given' do
+        expect { submit }
+        # TO WHAT? Change question count by one? how make that happen?
```
dfaulken: Ah, I suspect this is the cause of that method being untested according to simplecov. I'm kind of surprised that `expect` without a `to` even passes.

I notice that your `submit` let block doesn't have any changes that should be made here. A good place to start would be not worrying about questions at all - just submit an update to an attribute of the draft, and make sure that this controller method does that change (I'm pretty sure this is all the corresponding test on ft-forms does).

dfaulken:
Good work besides my comments from above and the untested method. :+1:

spec/controllers/application_drafts_controller_spec.rb
dfaulken: Changed my mind, since mocking it will make it not work.

PR #93, opened 2015-09-08
Application template creation fix

Anbranin:
That last commit should do it. Everything seems to work.

### DIFF
--- a/app/controllers/application_templates_controller.rb
+++ b/app/controllers/application_templates_controller.rb
@@ -10,11 +10,9 @@ class ApplicationTemplatesController < ApplicationController
   end
 
   def new
-    # make a new empty template
-    # make a draft for that template and go to its edit page
     position = Position.find(params.require :position_id)
-    @template = ApplicationTemplate.create!(position: position)
-    @draft = @template.create_draft @current_user
+    template = ApplicationTemplate.create!(position: position)
+    @draft = template.create_draft @current_user
     redirect_to edit_draft_path(@draft)
   end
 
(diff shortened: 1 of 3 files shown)


app/controllers/application_templates_controller.rb
```diff
@@ -9,6 +9,15 @@ def bus
     render 'show'
   end
 
+  def new
+    # make a new empty template
+    # make a draft for that template and go to its edit page
```
dfaulken: I would say you can go ahead and remove these comments. Hopefully it's pretty simple to see what your controller method does just from looking at the code.

app/controllers/application_templates_controller.rb
```diff
@@ -9,6 +9,15 @@ def bus
...
+    position = Position.find(params.require :position_id)
+    @template = ApplicationTemplate.create!(position: position)
```
dfaulken: This shouldn't need to be an instance variable, right?

app/models/application_template.rb
```diff
@@ -8,7 +8,7 @@ class ApplicationTemplate < ActiveRecord::Base
   belongs_to :position
   delegate :department, to: :position
 
-  validates :position, presence: true, uniqueness: true
+  validates :position, presence: true
```
dfaulken: What's your thinking on removing that validation?

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 99.43
+    "covered_percent": 98.67
```
dfaulken: Don't forget to test your method.

PR #99, opened 2015-11-01
Unique prompts

Anbranin:
Gave prompts a unique name attribute that prevents them from being duplicates and thus erased due to being duplicate keys in the hash sent to the controller. Put a method in the application helper that turns the hash into an array and strips the unique name attribute from them so it can be displayed in human readable format in the view.
Rewrote tests, tests pass. Still need to write a test for the parse_application_data method that is now in there.

closes #99


app/models/question.rb
```diff
@@ -43,6 +43,14 @@ def takes_placeholder?
     %w(text date).include? data_type
   end
 
+  def unique_prompt_name
```
dfaulken: Do we need unique names and prompts if we're serializing in an Array?

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 99.62
+    "covered_percent": 43.58
```
dfaulken: :zap:

spec/lib/tasks/record_import_rake_spec.rb
```diff
@@ -0,0 +1,3 @@
+require 'rake'
```
dfaulken: Don't need to include this if you don't make any progress

app/models/question.rb
Anbranin: Not sure. We should discuss this.

Anbranin: So, params[:responses] is still a hash when it gets sent to the controller, even if afterwards it is saved as an array. I could be completely off base with this, but I believe the duplicate keys will still get erased on the way to the controller.

dfaulken:
It occurs to me that I might have been wrong about developing the export task (CSV view) in this branch.

Anbranin:
I can roll-back-or-whatever-it's-called all the relevant commits. We should start working on this again soon.

app/helpers/application_helper.rb
```diff
@@ -6,6 +6,25 @@ def configured_organization_name
     configured_value [:organization_name]
   end
 
+  def parse_application_data(data)
+    prompts = []
+    responses = []
+    p_r = []
```
dfaulken: wuzzat

app/models/application_record.rb
```diff
@@ -6,16 +6,18 @@ class ApplicationRecord < ActiveRecord::Base
   delegate :department, to: :position
   has_one :interview, dependent: :destroy
 
-  serialize :responses, Hash
+  serialize :responses, Array
 
   validates :position,
-            :responses,
+            # :responses, put back in once import is complete
```
dfaulken: ?

app/models/question.rb
dfaulken: Correct.

spec/models/application_record_spec.rb
```diff
@@ -50,6 +71,45 @@
     end
   end
 
+  describe 'add_response_data' do
+    context 'application record with no existing responses' do
+      before :each do
+        headers = %w(x y z)
```
dfaulken: Perhaps a comment here explaining that the headers don't matter

spec/models/application_record_spec.rb
```diff
@@ -50,6 +71,45 @@
...
+      let :call do
+        @record.add_response_data(@question, @answer)
+      end
+      it 'makes the changes to the correct application record' do
+        expect(call.responses).to eql [[@question, @answer]]
+      end
+    end
+    context 'application record with existing responses' do
+      before :each do
+        @question1 = 'question1'
```
dfaulken: Maybe names like '@old_question` and `@new_question` would be more descriptive

dfaulken:
Opened separate PR with my test fix in https://github.com/umts/jobapps/pull/103.

PR #101, opened 2015-11-13
Csv export

Anbranin:
I hope everything is there

### DIFF
--- a/app/controllers/application_records_controller.rb
+++ b/app/controllers/application_records_controller.rb
@@ -4,7 +4,7 @@ class ApplicationRecordsController < ApplicationController
 
   def create
     create_user if @current_user.blank?
-    params.require(:responses)
+    params.require :responses
     params.require :position_id
     ApplicationRecord.create(position_id: params[:position_id],
                              responses: params[:responses],
--- a/app/controllers/application_records_controller.rb
+++ b/app/controllers/application_records_controller.rb
@@ -4,7 +4,7 @@ class ApplicationRecordsController < ApplicationController
 
   def create
     create_user if @current_user.blank?
-    params.require :responses
+    params.require(:responses)
     params.require :position_id
     ApplicationRecord.create(position_id: params[:position_id],
                              responses: params[:responses],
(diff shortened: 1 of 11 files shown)


app/controllers/application_records_controller.rb
```diff
@@ -1,10 +1,10 @@
 class ApplicationRecordsController < ApplicationController
   skip_before_action :access_control, only: [:create, :show]
-  before_action :find_record, except: :create
+  before_action :find_record, except: [:create, :csv_export]
 
   def create
     create_user if @current_user.blank?
-    params.require :responses
+    params.require(:responses)
```
dfaulken: ok

app/models/application_record.rb
```diff
@@ -9,13 +9,15 @@ class ApplicationRecord < ActiveRecord::Base
   serialize :responses, Hash
 
   validates :position,
-            :responses,
+            :responses, # this validation makes the test fail
```
dfaulken: :alarm_clock:

spec/controllers/application_records_controller_spec.rb
```diff
@@ -47,6 +53,28 @@
     # No reason for staff to use this page
   end
 
+  describe 'GET #csv_export' do
+    before :each do
+      when_current_user_is :staff
+      @start_date = 1.week.ago.to_date
+      @end_date = Time.zone.today
```
dfaulken: We should look into *why* Rubocop objects to `Date.today`.

spec/controllers/application_records_controller_spec.rb
```diff
@@ -137,6 +165,7 @@
       it 'assigns the correct variables' do
         submit
         expect(assigns.keys).to include 'record', 'interview'
+        # why interview?
```
dfaulken: I bet you can figure out the answer to that question

spec/views/application_records/csv_export_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'rails_helper'
+require 'csv'
+include RSpecHtmlMatchers
+
+describe 'application_records/csv_export.csv.erb' do
+  before :each do
+    @id = 1
```
dfaulken: Seeing hardcoded IDs makes alarms go off in my head. It's fine here!

But perhaps a quick comment explaining the structure of this test and the motivation thereof (since it's the only one of its kind).

spec/controllers/application_records_controller_spec.rb
Anbranin: http://rails-bestpractices.com/posts/2014/10/22/use-time-zone-now-instead-of-time-now/

PR #103, opened 2015-11-13
Unique prompts

dfaulken:
@Anbranin's original PR, with a test fix I wrote.


dfaulken:
Here's the plan for Sunday:

1. Perform a CSV export.
1. Set all existing application records to have `nil` responses.
1. Merge this branch and deploy.
1. Execute the CSV import task with the exported CSV file.

And then some spot-checks to make sure everything looks alright (I have some concerns about the character formatting of the CSV export).

PR #105, opened 2015-11-16
Removing blanks from new app records by including the headings and ex…

dfaulken:
…planations

No change to coverage %.


dfaulken:
Need to resolve merge conflict in ApplicationRecordsController

PR #111, opened 2015-12-02
Bump Ruby version and update gems

dfaulken:
Pessimistic version control on Rails gem.

Received deprecation warning about calling _path from a mailer, which is
good since that didn't make sense to do anyway. Changed to _url and then
needed to update the environment configs to account for this. If we
intend to send email with inward-pointing URLs in the future in other
apps, we will need to make the same change.


werebus:
Whoops, I mis-represented how pessimistic versioning works in yesterday's code review.  What we want is `~> 4.2`.  That is, "greater than 4.2, but not 5.x".  `~> 4`  turns out to be meaningless.  Sorry about that.

https://robots.thoughtbot.com/rubys-pessimistic-operator

Issue #104, opened 2015-11-15
Lingering issues from CSV export/import


dfaulken:
1. The file was comma-separated but some of the prompts had commas in them. The import threw out any columns after the first two, so we've lost that data (on the site, some manual editing of the CSV file could salvage this);
2. There are some character formatting weirdnesses (all apostrophes, quotation marks, etc.), and all question marks seem to have vanished?

Example:

![screen shot 2015-11-15 at 1 28 04 pm](https://cloud.githubusercontent.com/assets/3988134/11170260/c4b8e768-8b9c-11e5-8b26-988063d0e535.png)
![screen shot 2015-11-15 at 1 28 18 pm](https://cloud.githubusercontent.com/assets/3988134/11170261/c4ba10ac-8b9c-11e5-9d6c-2db6f288311f.png)

werebus:
On number 1: kind of late in hindsight, but ruby's [`CSV` library][1] has a `to_csv` method included into `Array` that will do the work of figuring out escaping for you to avoid that problem.

On number 2: `&#39;` *is* a single-quote, and it's stored as `&#39;` in the database for some reason, but Rails is being helpful and escaping it for you, so it comes to the browser as `&amp;#39;`

[1]:   http://ruby-doc.org/stdlib-2.0.0/libdoc/csv/rdoc/CSV.html

dfaulken:
Thanks @werebus - it couldn't hurt to go ahead and make that change to the CSV export (since it is something people may wish to use in the future).

enragednuke:
So is this closed? I see merged PRs that reference this but this is still open.

werebus:
Not really, no.  The bug that caused the first half of this issue is fixed. However, this:

>  ...The import threw out any columns after the first two, so we've lost that data (on the site, some manual editing of the CSV file could salvage this)...

hasn't been addressed as far as I know.  The second half of this issue is likely a wontfix.

enragednuke:
I'm confused, based on the page Dave edited, there seems to be only 3 columns (record id, prompt, response). I see prompt and response in the screenshot, so I assume it isn't losing any columns (I'm also assuming the record ID just isn't shown)

Do you mean rows? I see in the screenshot there's only the first two rows. It seems like a big issue not to fix.

werebus:
So the screenshot doesn't show an example of issue 1.  Here's a quick (made-up) example though

```csv
record_id, prompt, response
1, List 3 things you like, Well, I like puppies, kittens, and flowers
```

[...]

PR #113, opened 2015-12-13
Using #to_csv in CSV export

dfaulken:
Addresses #104. No change to coverage %


dfaulken:
Hub pulled the wrong info. Closing

PR #117, opened 2015-12-17
Interview can now be marked as hired or not.

### LINKED ISSUE #112: Hired or not?
This came up in #100, that although we have a `hired` attribute on `Interview`, it doesn't look like there's currently a way to mark an interview one way or the other.

Anbranin:
Resolves #112 

Here is what the buttons look like:
screen shot 2015-12-17 at 4 54 55 pm

That part of the site is not very pretty in general, so I definitely want to do some styling to make it better, but want that to be its own thing so I just tried to make it look passable for now.

I ended up making another attribute on Interview so people (Don) could make notes about the interview if the person wasn't hired, made two buttons and controller logic to handle them, displayed more in the past application page, and wrote some tests for (I think) everything. As usual, let me know what I can improve upon!

### DIFF
--- a/app/views/application_records/past_application_records.haml
+++ b/app/views/application_records/past_application_records.haml
@@ -1,11 +1,11 @@
 %table.data_table.display
   %thead
     %tr
-      %th= 'Date Submitted'
-      %th= 'Name'
-      %th= 'Interview'
-      %th= 'Staff Note'
-      %th= 'Hired?'
+      %th Date Submitted
+      %th Name
+      %th Interview
+      %th Staff Note
+      %th Hired?
   %tbody
     - @records.each do |record|
       %tr
@@ -25,7 +25,7 @@
           = record.staff_note
         %td
           - unless record.interview.nil?
-            - if record.interview.hired == true
+            - if record.interview.hired?
               Yes
             - else
               No -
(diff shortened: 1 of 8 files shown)


app/views/application_records/past_application_records.haml
```diff
@@ -5,6 +5,7 @@
       %th= 'Name'
       %th= 'Interview'
       %th= 'Staff Note'
+      %th= 'Hired?'
```
dfaulken: While we're here let's fix these `th`s to not needlessly use Ruby.

app/views/application_records/past_application_records.haml
```diff
@@ -22,3 +23,10 @@
             = format_date_time record.interview.scheduled
         %td
           = record.staff_note
+        %td
+          - unless record.interview.nil?
+            - if record.interview.hired == true
```
dfaulken: It's fine (and better) just to say

```ruby
if record.interview.hired?
```

spec/controllers/interviews_controller_spec.rb
```diff
@@ -19,9 +19,36 @@
...
-      it 'marks interview as complete' do
-        submit
-        expect(@interview.reload.completed?).to eql true
+      context 'hired button is pressed' do
+        let :submit do
+          post :complete, id: @interview.id, hired: true
+        end
+        it 'marks interview as complete' do
+          submit
+          expect(@interview.reload.completed?).to eql true
```
dfaulken:
```ruby
expect(@interview.reload).to be_completed
```

works too.

PR #124, opened 2015-12-23
require CSV for the application

werebus:
so out tests don't lie to us and @sherson doesn't look foolish in a demo


sherson:
Wasn't a demo (I was just trying it out) but thanks!

PR #127, opened 2015-12-24
Site texts and app display edit

### LINKED ISSUE #119: Modify headings and expanations on show page
Don:

>  Good Morning,

>  The online application seems to have had an issue overnight. The Job Requirements portion has gone Bold.

---

@sherson

>  I talked to Don... everything is mostly fine. 

>  What he's referring to is that all prompts and responses appear when viewing recent applications ~~(that were submitted since the last code commits)~~. 

>  ~~When viewing old applications, the text for headings, explanations, and one of the yes/no fields aren't rendered (the rest of the yes/no fields render).~~

>  The change is especially noticeable since the "Minimum Job Requirement..." text is long (and that column is bold). It's a tad ugly rendered as a wall instead of a list, but the new behavior is what I'd expect.

---

@sherson:

>  Never mind what I wrote about a yes/no field not appearing. I just misread the page (only the headings and explanations weren't showing).

>  Anyway, it's all set.

Anbranin:
Closes #119 

Application record show page now contains any markdown used in application drafts/template. What the applicant sees is now what we see on the application record display page.

Tests written for new site texts controller logic that will display a 404 not found error when the site text does not exist. Also, if there is no site text, the link to the markdown explanation will not display.

### DIFF
--- a/app/views/application_drafts/edit.haml
+++ b/app/views/application_drafts/edit.haml
@@ -1,3 +1,8 @@
+Note: The data type of 'explanation' supports markdown.
+- if @site_text.present?
+  For a quick tutorial, please click
+  = link_to 'here.', site_text_url(@site_text.name) 
+
 %h1.title Editing #{@draft.position.name_and_department}
 #flex_wrapper
   .form_column
--- a/app/views/application_drafts/edit.haml
+++ b/app/views/application_drafts/edit.haml
@@ -1,7 +1,7 @@
 Note: The data type of 'explanation' supports markdown.
 - if @site_text.present?
   For a quick tutorial, please click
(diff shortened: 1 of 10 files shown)


app/views/application_drafts/edit.haml
```diff
@@ -1,3 +1,8 @@
+Note: The data type of 'explanation' supports markdown.
+- if @site_text.present?
+  For a quick tutorial, please click
+  = link_to 'here.', site_text_url(@site_text.name)
```
dfaulken: I'd also have been fine with just writing

```ruby
site_text_url('markdown explanation')
```

and leaving out the instance variable, but what you've done is fine.

dfaulken:
Looks good.

PR #130, opened 2015-12-27
Add custom RSpec matcher for requests that redirect back

dfaulken:


werebus:
Nifty.  Reads much more nicely.

PR #131, opened 2015-12-27
Refuse coverage percentage drops

dfaulken:
This will produce failure messages locally, but the most important
consequence will be Travis builds failing outright.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -3,7 +3,9 @@ require 'factory_girl_rails'
 require 'simplecov'
 
 CodeClimate::TestReporter.start if ENV['CI']
-SimpleCov.start 'rails'
+SimpleCov.start 'rails' do
+  refuse_coverage_drop
+end
 
 RSpec.configure do |config|
   config.raise_errors_for_deprecations!


werebus:
While I like the general idea and always-increasing code coverage is an admirable goal, perhaps we could go with something just a smidge [more forgiving][1]?

[1]:  https://github.com/colszowka/simplecov#maximum-coverage-drop

dfaulken:
Well, I suppose it depends what kind of development scenarios we want to account for. Between the two ways of allowing some slip, I would prefer setting a minimum overall percentage to allowing incrementally worse coverage over time.

But on the other hand, under what circumstances would you suggest coverage percentage be allowed to drop?

Anbranin:
When we chatted about this, @werebus said the two cases where the coverage percentage dropping would be okay would be if the coverage percentage only dropped because of the way things were measured (could happen by a tiny percentage, not likely but still could happen), or what if we really wanted to get something out quickly with the intention of testing it later? But those seem like edge-case arguments and I can't see the harm, either, in trying your suggestion. Not like we can't reverse it if we find that there's something we haven't thought of. I must say for the record though that @werebus's suggestion wins my heart.

werebus:
Yeah, I guess you're right.

I think what I'm actually feeling is that there are different kinds of failure and they're all wrapped up in a `1` exit-code. When it comes down to it, code coverage just isn't as important as spec failures, but there isn't a way to offer those gradations in Travis.

PR #133, opened 2015-12-29
Take down apps

### LINKED ISSUE #74: Take down applications
ST requested that they be able to take down applications for when they're not hiring.

Anbranin:
What do you all think? I left off the idea of doing anything to application drafts, and instead added an 'active' boolean attribute (default active) to the application template. I then added a little snippet to the application template that looks like this:
screen shot 2015-12-29 at 4 19 19 pm
When the application is active, and this:
screen shot 2015-12-29 at 4 17 45 pm
when the application is marked as inactive.
For the student, the link to the inactive application template does not show on their dashboard. All active templates are shown.
closes #74


app/views/dashboard/student.haml
```diff
@@ -21,5 +21,6 @@
       - if app_submitted && should_show
         %li= link_to "Review your application for #{position.name}", application
       - elsif position.application_template.present?
-        %li= link_to "Submit application for #{position.name}",
-          application_template_path(position.application_template)
+        - if position.application_template.active?
```
werebus:
```ruby
elseif postition.application_template.present? && position.application_template.active?
```
perhaps? (although it moight be too long)

spec/views/dashboard/student_spec.rb
```diff
@@ -100,6 +101,17 @@
...
       end
+      context 'applications for the position are inactive' do
+        before :each do
+          @template = create :application_template,
+                             position: @position,
+                             active: false
+        end
+        it 'does not list the link to submit application for position' do
+          render
+          expect(rendered).not_to have_tag 'a'
```
werebus: No `<a>` tags at all?  Seems like this might not be specific enough.

app/views/application_templates/show.haml
```diff
@@ -2,6 +2,21 @@
   The #{@template.position.name_and_department} application
   is not currently available.
 - else
+  - if @current_user.try :staff?
+    - if @template.active?
+      Not hiring currently? Click here to take down the application:
```
werebus: "Click here" vs "Click below"

Also, tone here is a little different than the rest of the site (not that I mind, per se)

app/controllers/application_templates_controller.rb
```diff
@@ -19,6 +19,15 @@ def new
   def show
   end
 
+  def toggle_active
```
werebus: Ultimately, this controller doesn't "toggle" active, it sets it depending on the value of `params[:up]` or `params[:down]`

app/views/dashboard/student.haml
werebus: Or maybe:

```ruby
elsif position.application_template.try(:active?)
```

werebus:
One last suggestion:

When there aren't any active applications, the student dashboard looks a little odd

screenshot_2015-12-30_11-41-34

Some "Sorry, we don't want you right now" wording, perhaps?

Anbranin:
What about "No applications are currently available. Please try again later."

werebus:
"Positions", perhaps?

Anbranin:
I came up with something, see if you all like it. @sherson agreed that it was desirable behavior to list the position that was not hiring with an announcement.

app/controllers/application_templates_controller.rb
Anbranin: Oooh I have the perfect name.

sherson:
Looks great, but it's a little weird that staff can't activate/deactivate applications from the editor (would be convenient if they could do it from both the editor and viewer).

Anbranin:
https://www.google.com/url?sa=i&rct=j&q=&esrc=s&source=images&cd=&ved=0ahUKEwi9_vOXv4TKAhXGOCYKHXDWCz8QjRwIBw&url=http%3A%2F%2Fblog.stacksocial.com%2F6-giveaways-didnt-make-cut-thanksgiveaway%2F&bvm=bv.110151844,d.dmo&psig=AFQjCNEdC9gXLfYFzMr_-vHqhQiUfi5dPA&ust=1451596147709791

sherson:
darth vadar snow
(seems more appropriate today)

app/controllers/application_templates_controller.rb
```diff
@@ -19,6 +19,15 @@ def new
   def show
   end
 
+  def set_active
+    if params[:down]
+      @template.update!(active: false)
```
dfaulken: Why bang?

spec/controllers/application_templates_controller_spec.rb
```diff
@@ -109,4 +109,47 @@
...
+      before :each do
+        when_current_user_is :staff
+      end
+      context 'button to reactivate application template has been pressed' do
+        let :submit do
+          post :set_active, id: @template.id, up: 'value'
+        end
+        it 'marks the application as active' do
+          submit
+          expect(@template.reload.active).to eql true
```
dfaulken:
```ruby
expect(@template.reload).to be_active
```

should work too (you could even try it without the `reload` to see if that works).

app/controllers/application_templates_controller.rb
```diff
@@ -19,6 +19,15 @@ def new
   def show
   end
 
+  def set_active
```
dfaulken: I think I'd prefer to accomplish this in one of two ways:

1. Have a controller action which toggles the `active` attribute rather than having to specify it in a parameter - that way, you only need to write the link once, and you can make use of `ActiveRecord::Persistence#toggle`;
2. Instead of writing a controller action just for making DB changes, write a specialized form which just makes a PATCH request. The downside of this is that since the request will go to the `#update` controller action, it will redirect/render the same things the normal editing structure would.

I prefer the first solution.

Anbranin:
This is on the application drafts AND application template page, and goes to a new controller action #toggle_active:
screen shot 2016-01-05 at 10 58 57 am
screen shot 2016-01-05 at 10 59 05 am

In addition, this is the type of thing you'll see as a student if any applications aren't live:
screen shot 2016-01-05 at 10 58 28 am

Thoughts?

app/controllers/application_templates_controller.rb
Anbranin: Now it _literally_ toggles and is named toggle_active.

spec/views/application_drafts/edit_spec.rb
```diff
@@ -9,6 +9,27 @@
...
+      render
+      expect(rendered).to have_form action_path, :post do
+        with_tag :input, with: { type: 'submit',
+                                 value: 'Deactivate application' }
+      end
+    end
+  end
+  context 'application template is inactive' do
+    it 'has a button to reactivate the application' do
+      @template.update active: false
```
dfaulken: Rather than a lot of tests that update pre-existing objects, perhaps a separate pre-defined object?

spec/views/dashboard/student_spec.rb
```diff
@@ -96,8 +95,21 @@
       end
       context 'no applications are present for position' do
         it 'does not list links to submit application for position' do
+          @template = create :application_template,
```
dfaulken: Does this need to be an instance variable (same on line 107)?

app/controllers/application_templates_controller.rb
```diff
@@ -19,6 +19,11 @@ def new
   def show
   end
 
+  def toggle_active
+    @template.toggle!(:active)
```
dfaulken: Why bang?

spec/views/application_drafts/edit_spec.rb
Anbranin: What is the difference between a 'pre-existing' object and a 'pre-defined' one?

spec/views/dashboard/student_spec.rb
Anbranin: Ah... nope :)

app/controllers/application_templates_controller.rb
Anbranin: Because bang saves the updated object, "toggle" by itself does not. I _think_ I would have to return self if I did not use bang.

Anbranin: Or rather I'd have to call .save on the toggled object. derp

spec/views/application_drafts/edit_spec.rb
dfaulken: I'm saying instead of creating an object and then later modifying it before you render a view, it's better (and easier to read) if you create all the objects you need beforehand.

So in the `before` block:

```ruby
@active_template = create :application_template, active: true
@inactive_template = create :application_template, active: false
assign :template, @active_template
```

And then in a test where you want to render the active template:

```ruby
assign :template, @inactive_template
```

app/controllers/application_templates_controller.rb
dfaulken: Oh, interesting. That's different from the usual Rails convention for what a bang is - I would've expected an exception instead of a silent failure (à la `update` vs. `update!`). I agree with your thinking.

spec/views/application_drafts/edit_spec.rb
dfaulken: I also notice that you're not differentiating between test instance variables and view instance variables - don't forget to use `assign` to set view instance variables.

app/views/dashboard/student.haml
```diff
@@ -20,6 +20,10 @@
...
       - if app_submitted && should_show
         %li= link_to "Review your application for #{position.name}", application
-      - elsif position.application_template.present?
+      - elsif position.application_template.try(:active?)
         %li= link_to "Submit application for #{position.name}",
           application_template_path(position.application_template)
+      - else
+        %li
+          We are not currently hiring for the position,
+          #{position.name}. Please check back.
```
dfaulken: @sherson, does this text seem okay to you?

spec/controllers/application_templates_controller_spec.rb
```diff
@@ -109,4 +109,50 @@
...
+        request.env['HTTP_REFERER'] = 'http://test.host/redirect'
+      end
+      context 'button to activate application template has been pressed' do
+        let :submit do
+          post :toggle_active, id: @template.id
+        end
+        it 'marks the application as active' do
+          @template.update active: false
+          submit
+          expect(@template.reload).to be_active
```
dfaulken: could even go with

```ruby
expect { submit }.to change { @template.reload.active? }

dfaulken:
Tests are really excellent - this is good to go once we confirm the student-facing text.

app/views/dashboard/student.haml
dfaulken: :zap:

dfaulken:
Do me a favor - please pull from `origin/master` so that the updated Travis YML file will give this PR an accurate build status.

app/views/dashboard/student.haml
sherson: :+1:

PR #137, opened 2016-01-04
Filter by dept

### LINKED ISSUE #126: Exports / reports by position
At the moment all of the exports and reports on the dashboard don't discriminate by department.  Probably they should be filterable or something.

Anbranin:
Filtering by department now works. Thoughts? I added a new plugin.. perhaps this is undesirable. But it looks so pretty, you guys. LOOK AT IT
screen shot 2015-12-30 at 4 04 59 pm

closes #126 

To clarify, I know the issue title states that the filtering should occur by position, but this is in fact not what should occur. It should filter by department.


app/assets/javascripts/sumoselect.js.coffee
```diff
@@ -0,0 +1,4 @@
+$(document).ready ->
+  $('.dept_select').SumoSelect(
+    placeholder: 'Select department')
+  return
```
dfaulken: One line would be fine, and it'd be good if it indicated that one can select multiple departments:

```coffee
$(document).ready ->
  $('.dept_select').SumoSelect placeholder: 'Select department(s)'
  return
```

app/models/application_record.rb
```diff
@@ -19,6 +19,11 @@ class ApplicationRecord < ActiveRecord::Base
   scope :newest_first, -> { order 'created_at desc' }
   scope :between,
         -> (start_date, end_date) { where created_at: start_date..end_date }
+  scope :in_department,
+        lambda { |department_id|
```
dfaulken: I know Rubocop corrects `->` to `lambda`, but I disagree. We can change this one back to look like the other AR scopes.

app/views/dashboard/_eeo_data.haml
```diff
@@ -1,6 +1,11 @@
 .date_range
   = form_tag eeo_data_application_records_path, method: :get do
-    List EEO data between
+    List EEO data for
+    .note_item
+      = select_tag :department_id,
+      options_from_collection_for_select(@departments, :id, :name),
+      class: 'dept_select', multiple: true
```
dfaulken: Since we can select multiple department IDs, it'd be nice if the various `department_id` variable names here made that clearer - even just `department_ids`, I think, would be preferable.

spec/models/application_record_spec.rb
```diff
@@ -206,56 +238,74 @@
       # technically. gender_eeo_data uses it.
       call
     end
+    it 'calls AR#in_department to filter application records' do
+      relation = ApplicationRecord.all
+      expect(ApplicationRecord).to receive(:in_department)
+        .with(@department.id)
+        .and_return(relation).at_least(:once)
+      # needs to return an activerecord relation. Also
```
dfaulken: This comment could be worded more clearly, and should also go in the other places where we mock in `ApplicationRecord.none`.

spec/models/application_record_spec.rb
```diff
@@ -206,56 +238,74 @@
...
+        .with(@start_date, @end_date, @department.id)
         .and_return('something')
       call
       expect(call[:genders]).to eql 'something'
     end
     it 'puts all genders and the counts thereof in the hash' do
       create :application_record,
-             gender: 'Female'
+             gender: 'Female',
+             position: @position
```
dfaulken: In cases where a bunch of things are the same except for one attribute, I like to put just that attribute on its own line, e.g.:

```ruby
create :application_record, position: @position,
  gender: 'Female'
```

app/models/application_record.rb
```diff
@@ -58,23 +63,24 @@ def pending?
...
     all_genders.map do |option|
       [option, gender_records.where(gender: option).count]
     end
   end
 
-  def self.eeo_data(start_date, end_date)
+  def self.eeo_data(start_date, end_date, department_id)
     records = {}
-    records[:all] = between(start_date, end_date)
+    records[:all] = between(start_date, end_date).in_department(department_id)
```
dfaulken: I'd prefer to leave off the final parentheses here - I think it reads better as a prepositional phrase that way.

app/controllers/application_records_controller.rb
```diff
@@ -22,14 +22,18 @@ def create
...
     render 'csv_export.csv.erb', layout: false
   end
 
   def eeo_data
     start_date = parse_american_date(params.require :eeo_start_date)
     end_date = parse_american_date(params.require :eeo_end_date)
-    @records = ApplicationRecord.eeo_data(start_date, end_date)
+    params.require :department_id
+    @records = ApplicationRecord
+               .eeo_data(start_date, end_date, params[:department_id])
```
dfaulken: Stylistically I think I prefer the look of line-splitting the params rather than like this.

app/models/application_record.rb
Anbranin: No. Rubocop will not have it.

dfaulken: We decided that `lambda` with block variables is better for multiline closures.

dfaulken:
Looks good.

PR #141, opened 2016-01-06
iCal Export Description

akaplo:
Closes #68.  I added the “description” property that’s included in the ical spec.  When a new export is created, it should populate Description with the absolute URL of the ApplicationRecord.  Does not include view tests, but will tomorrow.

PR #142, opened 2016-01-07
Add DESCRIPTION field to iCal export.

### LINKED ISSUE #68: Add info to calendar export
Staff asked for a phone number.  I think a link to the `ApplicationRecord` might be appropriate.

```
4.8.1.5 Description

   Property Name: DESCRIPTION

   Purpose: This property provides a more complete description of the
   calendar component, than that provided by the "SUMMARY" property.

   Value Type: TEXT

   Property Parameters: Non-standard, alternate text representation and
   language property parameters can be specified on this property.

   Conformance: The property can be specified in the "VEVENT", "VTODO",
   "VJOURNAL" or "VALARM" calendar components. The property can be
   specified multiple times only within a "VJOURNAL" calendar component.

   Description: This property is used in the "VEVENT" and "VTODO" to
   capture lengthy textual decriptions associated with the activity.

   This property is used in the "VJOURNAL" calendar component to capture
   one more textual journal entries.

   This property is used in the "VALARM" calendar component to capture
   the display text for a DISPLAY category of alarm, to capture the body
   text for an EMAIL category of alarm and to capture the argument
   string for a PROCEDURE category of alarm.

   Format Definition: The property is defined by the following notation:

     description        = "DESCRIPTION" descparam ":" text CRLF

     descparam  = *(

                ; the following are optional,
                ; but MUST NOT occur more than once

                (";" altrepparam) / (";" languageparam) /

                ; the following is optional,
                ; and MAY occur more than once

                (";" xparam)

                )


   Example: The following is an example of the property with formatted
   line breaks in the property value:

     DESCRIPTION:Meeting to provide technical review for "Phoenix"
       design.\n Happy Face Conference Room. Phoenix design team
       MUST attend this meeting.\n RSVP to team leader.

   The following is an example of the property with folding of long
   lines:

     DESCRIPTION:Last draft of the new novel is to be completed
       for the editor's proof today.

akaplo:
Closed PR #141 in favor of this.  Closes #68.  I added a description to the interview iCal export that contains the absolute url to the ApplicationRecord.  

As @werebus noted in the other PR, we may include more to make the field more helpful.  The interviewee's name is included, so perhaps we can include their email address as well.  #68 said staff asked for a phone number, but jobapps doesn't require a phone number from the interviewee.

There is currently no view test for the iCal export, so I'm going to create it in <code>spec/views/interviews</code> and base it off of <code>spec/views/application_records/csv_export_spec.rb</code>

### DIFF
--- a/app/views/interviews/interview.ics.erb
+++ b/app/views/interviews/interview.ics.erb
@@ -5,6 +5,7 @@ TZID:US-Eastern
 METHOD:PUBLISH
 BEGIN:VEVENT
 DTSTART:<%= format_date_time @interview.scheduled, format: :iCal %>
+DESCRIPTION:<%= application_record_url %>
 SUMMARY:<%= @interview.calendar_title %>
 UID:INTERVIEW<%= @interview.id -%>@UMASS_TRANSIT//JOBAPPS
 DTSTAMP:<%= format_date_time DateTime.now, format: :iCal %>
--- a/app/views/interviews/interview.ics.erb
+++ b/app/views/interviews/interview.ics.erb
@@ -5,7 +5,7 @@ TZID:US-Eastern
 METHOD:PUBLISH
 BEGIN:VEVENT
 DTSTART:<%= format_date_time @interview.scheduled, format: :iCal %>
-DESCRIPTION:<%= application_record_url(@interview.application_record_id) %>
+DESCRIPTION:<%= application_record_url(@interview.application_record) %>
 SUMMARY:<%= @interview.calendar_title %>
 UID:INTERVIEW<%= @interview.id -%>@UMASS_TRANSIT//JOBAPPS
 DTSTAMP:<%= format_date_time DateTime.now, format: :iCal %>
--- a/app/views/interviews/interview.ics.erb
+++ b/app/views/interviews/interview.ics.erb
@@ -5,7 +5,7 @@ TZID:US-Eastern
 METHOD:PUBLISH
 BEGIN:VEVENT
 DTSTART:<%= format_date_time @interview.scheduled, format: :iCal %>
-DESCRIPTION:<%= application_record_url %>
+DESCRIPTION:<%= application_record_url(@interview.application_record_id) %>
 SUMMARY:<%= @interview.calendar_title %>
 UID:INTERVIEW<%= @interview.id -%>@UMASS_TRANSIT//JOBAPPS
 DTSTAMP:<%= format_date_time DateTime.now, format: :iCal %>
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 42.98
+    "covered_percent": 99.83
   }
 }
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 99.83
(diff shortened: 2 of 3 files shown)


werebus:
Yeah, it's not practical to include anything that's a `Question` since there's no good way to guarantee the structure of that data.  Email address *might* be worthwhile, I dunno.

akaplo:
Gotcha.

I'm working on the view tests now and I've got a problem (this may be a @dfaulken question):

<code>ApplicationRecord</code>'s <code>show.haml</code> contains the following line:

<code>.export_link.note_item= link_to 'Click here for a calendar export file',
        interview_path(@interview, format: :ics)</code>, 

which I assume calls the Interviews controller's <code>show</code> method and passes it (the <b>@interview</b> object and a <b>format</b> property set to <i>ics</i>) as parameters.

This, in turn, renders the view <code>interviews/interview.ics.erb</code>, the nature of which, as described by the icalendar standard, is to return a .ics file with the properties described in <b>@interview</b>.

For my tests, I want to check that the properties of the .ics file contain the correct information.  The way to do this in rspec, I assume, is by creating a Factory <b>@interview</b> object and calling <code>show(@interview, format: ics)</code> in <code>before :each</code>.
Then, in the tests, I'd call <code>render</code> somehow check that the rendered html contains the right values for each property.  

Google hasn't been too helpful with this, because everyone else uses a gem called icalendar and doesn't render ics files as views.

dfaulken:
Correct, although you wouldn't be making assertions about the rendered *HTML* but rather about the rendered ics file.

Also, maybe we should look into that gem. I didn't know there was one when I wrote the export.

akaplo:
Okay, then there's somewhere small that I'm going wrong.  When I call <code>show(@interview, format: :ics)</code>, I'm getting 
<i>
failure/Error: show(@interview, format: :ics)
     
     NoMethodError:
       undefined method `show' for #<RSpec::ExampleGroups::InterviewsInterviewIcsErb:0x007ff83f7498e0>
     # /Users/akaplowitz/.rvm/gems/ruby-2.2.3/gems/actionpack-4.2.5/lib/action_dispatch/testing/assertions/routing.rb:171:in `method_missing'
     # /Users/akaplowitz/.rvm/gems/ruby-2.2.3/gems/actionview-4.2.5/lib/action_view/test_case.rb:272:in `method_missing'
     # ./spec/views/interviews/ical_export_spec.rb:7:in `block (2 levels) in <top (required)>'
</i>

That obviously means I'm calling it wrong, but I don't know why :(

dfaulken:
Ah, no, I didn't read your comment quite carefully enough. View tests infer what file you'd like `render` to render [from a combination of the spec file name and the argument to the describe block](https://www.relishapp.com/rspec/rspec-rails/v/2-1/docs/view-specs/view-spec-infers-controller-path-and-action). You don't specify the desired action in any way other than that, and just calling `render` is what should do the trick.

akaplo:
Ah, I see.

Removing that line and simply calling <code>render</code> produces its own issue:
<code>
ActionController::UrlGenerationError:
     #   No route matches {:action=>"show", :controller=>"application_records"} missing required keys: [:id]
</code>

which is true.  It's supposed to call the <i>interview_controller</i>'s <code>show</code>, not <i>application_record_controller</i>'s.  I never specify ApplicationRecord anywhere, so I'm not sure why it thinks that's what I'm talking about..

dfaulken:
Let's move this discussion to Slack.

Issue #153, opened 2016-01-12
DRY up tests?


dfaulken:
When [I enabled linting of the spec directory for CodeClimate](https://github.com/umts/jobapps/commit/19668f18c1278d528440050373dd11f938d697e7), many of our tests received poor grades for duplication.

Do we consider this to be a problem? Should we invest in things like shared examples? Or do we think that modeling the same kind of abstraction in our tests that we have in the objects we're testing weakens our tests?

Should we add the `spec` directory back to [the exclude list](https://github.com/umts/jobapps/commit/19668f18c1278d528440050373dd11f938d697e7#diff-0655d430acfa953cae08b6619689d18cR28)?

dfaulken:
It's also worth noting that of the poorly-rated files, the majority are either view specs or controller specs. Perhaps a good chunk of this may be addressed through implementing #128.

PR #164, opened 2016-01-13
added an integration test for user deletion

bgregg:


dfaulken:
I've reviewed the codeclimate failure, and it seems fine. If anything, it's just a further symptom of #153.

dfaulken:
Looks gret.

PR #165, opened 2016-01-14
Add custom cop

dfaulken:
Use `be` matcher rather than `eq` or `eql` for singleton types
(boolean, fixnum, etc).

Has support for autocorrect, which I used here.

Next up: preferring predicate matchers to matching against the return value of a predicate method.

### DIFF
--- a/lib/custom_cops/be_for_singleton.rb
+++ b/lib/custom_cops/be_for_singleton.rb
@@ -12,6 +12,13 @@ module RuboCop
                                  int_type?
                                  float_type?)
 
+        def autocorrect(node)
+          lambda do |corrector|
+            matcher = node.child_nodes[1]
+            corrector.replace matcher.loc.selector, 'be'
+          end
+        end
+
         def on_send(node)
           return unless node.method_name == :to
           return unless node.child_nodes.first.method_name == :expect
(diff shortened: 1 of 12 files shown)


lib/custom_cops/be_for_singleton.rb
```diff
@@ -0,0 +1,36 @@
+module RuboCop
+  module Cop
+    module CustomCops
+      class BeForSingleton < Cop
```
dfaulken: Fully open to suggestions for better names.

werebus: Yeah, is "singleton" technically correct for `Fixnum`s and `Float`s?

dfaulken: No! I couldn't think of what it *is* called though.

werebus: "nonduplicable", perhaps?

https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/object/duplicable.rb

dfaulken: I like that better.

dfaulken:
Pulling this down in favor of getting umts/custom-cops working.

PR #166, opened 2016-01-14
deleted view spec for user and edited controller spec to reflect new …

bgregg:
…integration tests


dfaulken:
The codeclimate failure isn't an issue per #153.

PR #168, opened 2016-01-14
Department integration tests

bgregg:
^^RT

PR #169, opened 2016-01-14
Dual departments fix

### LINKED ISSUE #167: Multiple departments with the same name can be created
It seems to me that when someone creates a department that already exists, that department should not be able to be created. 

A simple validation should fix this.

bgregg:
Closes #167


dfaulken:
Looks like this PR includes the changes from #168 - be sure to check out to master before checking out to a new branch.

I'm closing this PR - please checkout to a fresh branch from master and open another one with only the relevant changes.

PR #172, opened 2016-01-14
Include umts-custom-cops

dfaulken:


dfaulken:
Weird that it caused codeclimate to fail.

werebus:
So, my suggestion for this is to make a codeclimate-specific `.rubocop.yml` file, I think.  Then tell codeclimate to use that.

PR #176, opened 2016-01-16
Change Travis notification settings

dfaulken:
Only notify on status change.

If we don't get an email for this PR's build, then this is probably what
we want.


dfaulken:
screen shot 2016-01-16 at 10 46 10 am

So we probably just want:

```yml
failure: change
success: never
```

PR #177, opened 2016-01-16
Change Travis email settings

dfaulken:
Never email on success; only email on failure if it's a change (don't send
successive emails for successive failing builds).


dfaulken:
We didn't get an email, so this seems correct. Merging

PR #171, opened 2016-01-14
Position Integration Tests

akaplo:
Added integration tests for Position and removed view/controller tests that aren't needed any longer. :+1:

PR #180, opened 2016-01-20
Spec helper method allows filling in select tags

### LINKED ISSUE #170: fill_in_fields_for doesn't work with foreign keys / select tags

dfaulken:
Closes #170.


dfaulken:
Merging since this affects #168 and #171.

PR #185, opened 2016-01-25
Change out sumo select

### LINKED ISSUE #183: Exception on past applications when no departments are selected
```
An ActionController::ParameterMissing occurred in application_records#past_applications:

  param is missing or the value is empty: department_ids
  app/controllers/application_records_controller.rb:45:in `past_applications'
```

Happens on past applications, but probably also on the other reports as well.  If the field is required, we should enforce that before the form gets submitted.

Anbranin:
HAHA Don't merge this yet.. please feel free to help me. Sumoselect is all gone and the other one is in its place, but it doesn't look right and I can't figure out how to make it look right.

Update: This looks good to merge unless anyone has issues.

Closes #183


Anbranin:
Okay so, I wrote my javascript incorrectly, and now it looks "right" meaning like this so far:
screen shot 2016-01-25 at 3 37 42 pm
Not sure why the placeholder isn't working, but at least I figured out where the initial problem was: I just forgot how to JS completely.

Anbranin:
This http://wenzhixin.net.cn/p/multiple-select/docs/#the-basics1

werebus:
I'd again like to make the case for [Rails Assets][1] over dropping things in `vendor` and forgetting about them.  External JS and CSS are software dependencies too.

[1]:  https://rails-assets.org/

app/views/dashboard/_past_applications.haml
```diff
@@ -5,7 +5,7 @@
       .note_item
         = select_tag :department_ids,
           options_from_collection_for_select(@departments, :id, :name),
-        class: 'dept_select', multiple: true
```
werebus: I think you should leave this.  I know we don't worry too much about progressive enhancement, but it looks like the plugin will work either way.

Anbranin: OK so I deleted the files in vendor, because you're absolutely right and I forgot about rails-assets. I followed the instructions on the rails asset site for how to install a gem that way and it looked like everything worked when I ran bundle.. BUT. 
screen shot 2016-01-25 at 4 11 26 pm
Did I miss a step? It said "installing rails-assets-multiple-select' and said bundle complete...

Anbranin: HAHA RESTARTING THE SERVER

Anbranin: @werebus everything is working now. Thank you for your patience in reteaching me things like 3 times before I get them maybe.

Anbranin:
Oh hey, guess what? This stupid plugin does the same thing as the other one. Required: True doesn't do anything, still, because this plugin probably also does the same "hiding" thing. I think the issue here is that I don't know how to tell before actually downloading it and trying it out. :dancers:

Anbranin:
K there's nothing that does what I want. Another thought: Make the field not required, and have the default be "all departments."

werebus:
A default of "all" is unlikely to be desirable.  I feel pretty sure that there is something out there that will do what you want.  This isn't all that weird a use-case.

That being said, the way form-validation behaves is pretty browser-specific.  On Firefox, with SumoSelect, `required=true`, I get a little box that says "Please select an item in the list", but it's positioned in the upper-left corner of the window.

Anbranin:
Decided to stay with multiple-select because I could get it from rails-assets, unlike sumoselect. This will be better. 
However, since it hides the initial DOM element, HTML validation wasn't showing up under all circumstances. Solution? Use a JQuery plugin that could validate hidden fields. I found one.  There is a lot to do including fixing tests, making it perhaps less ugly, and handling what happens when someone doesn't have Javascript enabled in their browser. The required: true attribute is still there, so HTML validation will work in that case, however it will not work on the hidden attribute and might get you a 500 error. This can be handled in the controller, no? Like "if params[:departments] == nil do this"

Anbranin:
Some refactoring definitely has to be done, although I'm not sure yet how to go about it. I'm at the point where it looks like everything's good and I just need others to look everything over and tell me what I've missed.

TODO
```diff
@@ -0,0 +1,10 @@
+[x] Make department filter not required
```
dfaulken: If you're in the habit of using a personal TODO file, that's fine - please put it in your personal .gitignore, particular for OS repos.

app/controllers/application_records_controller.rb
```diff
@@ -22,19 +22,27 @@ def create
   def csv_export
     start_date = parse_american_date(params.require :start_date)
     end_date = parse_american_date(params.require :end_date)
-    params.require :department_ids
-    @records = ApplicationRecord.in_department(params[:department_ids])
-               .between(start_date, end_date)
+    if params[:department_ids]
```
dfaulken: Some pretty similar logic between the three controller actions here, so a `before_action` is probably warranted.

app/views/dashboard/_eeo_data.haml
```diff
@@ -4,7 +4,7 @@
     .note_item
       = select_tag :department_ids,
       options_from_collection_for_select(@departments, :id, :name),
-      class: 'dept_select', multiple: true
+      class: 'dept-select', multiple: true
```
dfaulken: Interesting. What's the significance of this?

TODO
Anbranin: OH DARN. FORGOT

app/controllers/application_records_controller.rb
Anbranin: I know, I've been brainstorming ways to do it but the only thing I can see to abstract out is ApplicationRecord.in_department. The between(start_date, end_date) thing relies on parameters that are different in all 3 cases. Abstracting the "between" thing will only remove 2 lines from each controller because of how eeo_data works.
Do you have any thoughts?

app/views/dashboard/_eeo_data.haml
Anbranin: Underscores are not recognized by the javascript in this case.

dfaulken: Uh.. by which Javascript? jQuery? with underscores in classes? That doesn't seem correct. There may have been another issue in play.

app/controllers/application_records_controller.rb
dfaulken: Well, it seems like the general job of the logic you need is to come up with a list of department IDs to pass to your various methods based on the value of the `department_ids` parameter. Then you can just pass that list to the method in question. If you have further questions I'll be on Slack.

app/views/dashboard/_eeo_data.haml
Anbranin: Sure. But if I change the class name to have an underscore in it, it stops working.

dfaulken: Our class names generally do, so I'd prefer for this one to for consistency's sake. If you're having problems with that I'll be on Slack.

Anbranin: Okay, I was wrong, scratch that. Underscores versus dashes do not make a difference. I'm leaving it as a dash because it does not matter. 
I had changed it in the first place because I made the JS line a dash and the class name an underscore--it was a typo, and caused things not to work, and I was freaking out and then found the error.... changed it to a dash to make them match.

dfaulken: As I said before, I'd like for it to have an underscore since most of our other CSS class names have underscores. You should only need to make it match everywhere.

Anbranin: Oh, I didn't see that comment before I replied. Sure, consistency is good. Changing.

dfaulken:
This is good to merge, right?

Anbranin:
Yep, all set as long as someone has looked it over.

Issue #186, opened 2016-01-28
CodeClimate rubocop engine doesn't seem to support custom rubocop config files


dfaulken:
See https://github.com/codeclimate/codeclimate-rubocop/issues/39. If they did, then we could at least maintain a separate config file that doesn't require the custom matchers - although still blech, since it would be exactly two lines different from the original.

Considering dropping the CodeClimate integration from this app (and from the other three Rails 4 apps) if there isn't forward movement here.

dfaulken:
Disabling PR integration.

dfaulken:
Disabled PR integration and destroyed OAuth tokens as necessary for:

+ umts/jobapps
+ umts/ft-forms
+ umts/screaming-dinosaur
+ umts/detours
+ umts/custom-cops
+ umts/custom-matchers

PR #193, opened 2016-02-01
Changes to Application_Templates routing

### LINKED ISSUE #190: Route ApplicationTemplates as /applications
I think it's better looking. 

Could even route them by department name and position name also. An example might be `https://transit-jobapps.admin.umass.edu/applications/bus/operator`. Then we wouldn't have to have special controller actions (and DHH will give us a congratulatory :+1:).

We will want to take care to update the public website(s) as we roll out this change.

enragednuke:
* Now is `/applications/:id`
 * Can also find templates by using the format `/application/:department/:position`

Closes #190 

(send help I don't know how to work codeclimate)

(yes I know I pushed to my fork's master on accident)

### DIFF
--- a/app/controllers/application_templates_controller.rb
+++ b/app/controllers/application_templates_controller.rb
@@ -27,16 +27,24 @@ class ApplicationTemplatesController < ApplicationController
   private
 
   def find_template
-    unless params[:specific_path]
-      params.require :id
-      @template = ApplicationTemplate.find params[:id]
-    else
+    if params[:specific_path]
       params.require :department
       params.require :position
-      #I'm sure there's a more succinct way to do this with a command other than .find that returns AR::RNF if nothing is found
-      @template = ApplicationTemplate.find do |apptem| 
-        apptem.department.name == params[:department] && apptem.position.name == params[:position]
-      end || raise(ActiveRecord::RecordNotFound)
+      find_template_by_path
+    else
+      params.require :id
+      find_template_by_id
+    end
+  end
+
+  def find_template_by_id
+    @template = ApplicationTemplate.find params[:id]
+  end
+
+  def find_template_by_path
+    @template = ApplicationTemplate.find do |apptem|
+      apptem.department.name.casecmp(params[:department]) &&
+      apptem.position.name.casecmp(params[:position])
     end
   end
 end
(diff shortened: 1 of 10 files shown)


dfaulken:
Codeclimate is broken; see #186. If Travis passes you're good after someone looks at your code.

app/controllers/application_templates_controller.rb
```diff
@@ -27,7 +27,24 @@ def toggle_active
   private
 
   def find_template
-    params.require :id
+    if params[:specific_path]
```
dfaulken: Right, so this should be done via routing, not with controller method changes.

http://guides.rubyonrails.org/routing.html#dynamic-segments

enragednuke: I'm confused, isn't that exactly what I'm doing? But you need that additional parameter to be able to distinguish between WHICH route you're taking so you can specify which parameters to require. You can't just require all of them, because then both will fail to missing the other.

dfaulken: Yes, my bad. I misread your code. You don't need to preserve the original routing, which will simplify a lot of the code. You can assume it follows the department/position format - hence my comment in the issue about being careful to update the hardlink on the public site when we roll this out.

enragednuke: Gotcha. I'll swap this stuff out when I get in tomorrow.

enragednuke:
@dfaulken A bunch of tests exploded once I made it dept/pos only routing, but they're all fixed and this should be good to go unless there's an issue with any of the code as of now.

enragednuke:
Should've addressed all the issues mentioned with the latest push.

app/helpers/application_templates_helper.rb
```diff
@@ -0,0 +1,2 @@
+module ApplicationTemplatesHelper
```
dfaulken: Can delete this file

dfaulken:
Looks good - I'll take care of the file trim. Thanks for your work on this @enragednuke

enragednuke:
@dfaulken So I found a side-effect of this. This ended up affecting all the routes to toggle_active as well (since it uses `find_template` as well) so I need to edit all of the `toggle_active_application_template_url` links too and push that.

dfaulken:
Interesting - why wasn't this caught by tests?

enragednuke:
I don't think the tests were testing clicking on those links. I know I modified some tests in regards to it because they were failing when I modified RSpec.

PR #194, opened 2016-02-02
Expect Flash Message

### LINKED ISSUE #192: Use expect_flash_message for integration tests
Completely forgot about this helper method - it's nice not only to make sure that a method is there, but that it's being taken from the YML file.

akaplo:
Closes #192.  

Updated all integration tests to use the spec_helper method ```expect_flash_message``` for...well...expecting flash messages.  

Note that error messages remain unchanged, because they are rendered by a private ApplicationController method called ```show_errors```, and figuring out how to test for it has been determined to be out of the scope of this issue by @dfaulken.

### DIFF
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 46.55
+    "covered_percent": 99.84
   }
 }
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 99.84
+    "covered_percent": 46.55
   }
 }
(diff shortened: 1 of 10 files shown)


dfaulken:
Are there any controller tests where we could've used `expect_flash_message`?

akaplo:
Yes, there are plenty!  I thought #192 was just concerned with the current bunch of integration tests.

dfaulken:
Nope, my bad that that wasn't clear. Any place where we manually inspect the `flash` for its `message` content, it'd be better to replace it with the spec helper method - this includes controller tests.

akaplo:
No worries.  I'll change the controller tests tomorrow and update this PR then.

akaplo:
I lied yesterday when I said there are plenty of controller tests to change over.

All that I can find that weren't testing for flash **errors** are already using `expect_flash_message`

dfaulken:
Great, I'll look this over again but it should be good to go.

PR #197, opened 2016-02-04
Bootstrap Integration

enragednuke:
This is gonna be a big PR because of how many lines of code is in the Bootstrap JS/CSS libraries that have been installed (they're in vendor, alls good! just using the bootstrap gem alone was not sufficing)

I've also begun to do some formatting work on the main pages. Here's what the stuff I've changed looks like thus far (there's gonna be a good amount of stuff that is broken that I haven't touched though, since Bootstrap tends to implode standard non-Bootstrap formatting):

**A preview of what of I want to redesign the Staff Dashboard blocks as** (only one of them looks like this so far)
screen shot 2016-02-04 at 5 41 06 pm

**Application Template Preview**
screen shot 2016-02-04 at 5 42 10 pm

**Application Template Editting**
screen shot 2016-02-04 at 5 43 14 pm

Not ready to be pushed yet, but I wanted to put this up early for those to comment on what I've done thus far since this'll likely be a larger PR (not including bootstrap inflation).


dfaulken:
So I agree that adding the bootstrap source will be large - can we use rails_assets for this?

Once that's done, it would be preferable to make individual PRs for individual views. The larger PRs are, the less constructive feedback people will be able to give you on them.

enragednuke:
That's fine but I've already fixed two of the views (a lot of this will be simply fixing existing views, most of the 'new' content will be on the Staff Dashboard).

I don't have any experience with rails_assets, so I have no clue if we can use it for this. round-three just has it in its vendor folder (albeit, they don't have both the normal and the `.min` one, so I can remove one of them).

dfaulken:
Right, but fixing existing views is still a process that merits individual review of each atomic change. If you *can* PR two things separately, it is always preferable to, for this reason.

@werebus and I have both objected to dumping vendor assets into the vendor directory and calling it a day. In his words:
> External JS and CSS are software dependencies too.

Rails Assets helps us manage this in a much more maintainable way.

I'm closing this PR - look into rails_assets (which definitely does have bootstrap) and implementing it that way.

Then you can cherry-pick commits about other views into separate feature branches and create separate PRs.

PR #198, opened 2016-02-04
Test ApplicationRecordsController#create for staff

### LINKED ISSUE #155: Test ApplicationRecordsController#create for staff
A while ago, [I wrote the following comment](https://github.com/umts/jobapps/commit/a6d33f5b7669870f727a0deb98f0a8032050e794#diff-893e33fd61a43b402df3774a3d1feadaR28) about `ApplicationRecordsController#create`:

```ruby
# No reason for staff to use this page
```

Does this mean they can't? If so, this should be tested.
If they can, what should it do when they do? This should be tested.
Does it act just like it does if a student views it? If so, how do we know unless we test it?

dfaulken:
Closes #155.

### DIFF
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 99.84
+    "covered_percent": 99.83
   }
 }
--- a/spec/controllers/application_records_controller_spec.rb
+++ b/spec/controllers/application_records_controller_spec.rb
@@ -49,7 +49,22 @@ describe ApplicationRecordsController do
         expect(response).to redirect_to student_dashboard_path
       end
     end
-    # No reason for staff to use this page
+    context 'staff' do
+      before(:each) { when_current_user_is :staff }
+      it 'creates an application record as specified' do
+        expect { submit }
+          .to change { ApplicationRecord.count }
+          .by 1
+      end
+      it 'shows the application_receipt message' do
+        expect_flash_message :application_receipt
+        submit
+      end
+      it 'redirects to the student dashboard' do
+        submit
+        expect(response).to redirect_to student_dashboard_path
+      end
+    end
   end
 
   describe 'GET #csv_export' do


coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 99.84
+    "covered_percent": 99.83
```
enragednuke: DUN DUN DUNNNNNNNNNNNNNNN

dfaulken: https://github.com/umts/jobapps/commit/e061f5d96dceb01daeee69181c82a76c6387a8df

enragednuke:
Looks good if the tests are passing, though I feel like you should expect_flash_message *after* you submit (but I'm not sure how that works, so it's just kinda on-a-whim assumption, very easily could be wrong).

dfaulken:
[It's a mock](https://github.com/umts/jobapps/blob/master/spec/spec_helper.rb#L33-L37), so it has to go before. Agreed that it's slightly awkward - I tried to hint at that by putting `expect` in the method name.

enragednuke:
So I feel if it's a mock, one of two things are missing; if that tests it on its own, what's the point of submitting the form? (you don't test anything after it anyway, right?) If it's mocking it to occur with the submittal, should it not be checking for the flash_message afterward?

dfaulken:
Okay, so:

> if that tests it on its own, what's the point of submitting the form?
If we don't submit the form, nothing triggers the method call that the mock is testing for.

> If it's mocking it to occur with the submittal, should it not be checking for the flash_message afterward?
No. You have to establish the expectation that a certain sender will receive a certain message *beforehand* so that RSpec can set up an event listener - not after.

Try it yourself: switch the order and the test will fail.

enragednuke:
Hmm, aight I got it (I found the source for it). I just assumed that you had to have something to test afterward that is actually happened (my impression of mocks is that a certain call will cause a specific event to happen, not *test* for that specific event to happen).

I don't really think I'd call this a mock (referring to my prior comment of it being a coercing a program into  a specific result rather than acting normally) since the method is just waiting for a change in a certain something (I don't think I see any 'coercion' going on here, at least).

```
def expect_flash_message(name)
  expect_any_instance_of(ConfigurableMessages)
    .to receive(:show_message)
    .with(name, hash_including(:default))
end
```

dfaulken:
> I don't really think I'd call this a mock

**You** might not, but RSpec does: https://www.relishapp.com/rspec/rspec-mocks/docs

And Mocha does: http://gofreerange.com/mocha/docs/

These are the two stubbing and mocking frameworks we use.

enragednuke:
Oh that's word I'm looking for: stubs. Well, blasphemy anyhow. 'Mock' doesn't seem to fit an action that just sits and waits for something to change. 'Listener' feels much better. Oh well.

*shakes fist angrily*

PR #199, opened 2016-02-04
Bootstrap Integration (Staff Dashboard View)

enragednuke:
Using rails-assets instead of manually uploading the assets.

**A preview of what of I want to redesign the Staff Dashboard blocks as**
screen shot 2016-02-04 at 5 41 06 pm

Not ready to be pushed yet, but I wanted to put this up early for those to comment on the style incase it's not ideal for live.

There are some slight code changes to other views here, but nothing new. The changes are simply modifications to allow Bootstrap to render the views in the same regard they were before (though probably a different font).

The modifications to note are in the files [here](https://github.com/umts/jobapps/pull/199/files#diff-0ae8dbace6e67fd5a4550950bb0fbb18R1) and [here](https://github.com/umts/jobapps/pull/199/files#diff-8ed1aaf3e351d892b6c08abd858efdbfR1) (non-gem or minor Bootstrap-fix related changes)


dfaulken:
Looks nice, can we see an example with some more departments/positions?

enragednuke:
I'm not in tomorrow, so it'll have to wait until next week (I don't have round-three setup on my home computers) but yeah I'll make some and set it up when I'm in then.

dfaulken:
Sounds good, looks great so far! I'm excited about this

Anbranin:
So while we're talking about styling, Don came up to me and asked whether this:
screen shot 2016-02-05 at 10 26 42 am
could be different. He takes issue with the fact that the years go back so far. I don't have time to look into it right now, and it's definitely not an essential thing.. just thought I'd bring it up instead of ignoring Don. Thoughts?

werebus:
@Anbranin, that should probably go in its own issue.

Anbranin:
Yeah I thought it was tiny enough not to have its own issue, but it turns out it's more complicated than I had supposed.

enragednuke:
Alright, first question to anyone who knows departments, how long should I expect a department name to be at MAX? @umts/transit-programmers @umts/transit-read-only-programmers 

Because right now I've found a bug where this occurs with long department names:

screen shot 2016-02-08 at 2 08 40 pm

So I'm wondering if its just a matter of making that column slightly larger (I don't think the application portion will take up much width, but a good 30-40% whitespace is probably good for stylistic purposes and having a header column with width >30% would look odd I think).

Or perhaps should I find a way to display names that are longer in a different manner (wrapping comes to mind, but may not work for all names if there's one with a really long single word).

enragednuke:
Alright I did some Sass/SCSS magic and got this output:

screen shot 2016-02-08 at 2 57 02 pm

enragednuke:
A solution to the aforementioned problem would be lines like the following in the Department model

```ruby
validates :no_word_too_long, on: [:create, :update]
# . . .
def no_word_too_long
  longest_word = name.split.max_by(&:length)
  errors.add(:name, "cannot have a word over 14 characters long, please #{longest_word}") if longest_word.length > 14
end
```

Which would restrict all Departments from having a word in their name greater than 14 characters long.

dfaulken:
We definitely do not want to place database constraints because of UI implementation. There's nothing *inherent* about a department that means it can't have 15-letter words. The job of the UI is to accommodate the structure of the data, **not** the other way around/

dfaulken:
Field trip groups on round-three, for example, are mostly academic departments. Here's a quick query:

```ruby
FtGroup.pluck(:name).max_by{|name| name.split.map(&:length).max}
#=> "Athletics/Cheerleading"
```

And even if we take out slashes:

```ruby
FtGroup.where("name not like '%/%'").pluck(:name).max_by{|name| name.split.map(&:length).max}
#=> "Interdisciplinary Studies Institute"
```

"Interdisciplinary" is three letters longer than "Transportation" - and it's also worth pointing out that the number of letters only kind of matters since we're not using a monospace font.

Our page layout shouldn't put us in the position of worrying about how long the words are in our database.

enragednuke:
> We definitely do not want to place database constraints because of UI implementation. There's nothing inherent about a department that means it can't have 15-letter words. The job of the UI is to accommodate the structure of the data, not the other way around.

Of course, which is why it was only a suggestion. I wasn't planning on implementing it as my own ideal, but my choices have often been wrong so it didn't hurt to check, right?

This is all entirely why I'm asking for ideas on how to modify it. I'm fully aware that there may be departments (would there be applications for Departments with the name of FT groups though? I was taking mine from round-three's Department model) with longer names. 

My idea of reducing the text size based on length made it look odd, so I may just go back to a revamped version of multi-list format since in a style like the one I have now, there are indeed length restrictions (that can be fixed with font-size, but it looks weird when some of the page is 24pt and others are 18pt due to their length).

dfaulken:
My feeling is that if the presentation you're going for means that we have to obscure the meaning of our data under certain circumstances, then it's the wrong presentation format - in particular, that we should move away from trying to fit department names into a small column on the left if they won't fit there sometimes.

dfaulken:
Perhaps individual panels for each department, each with their own centered department name at the top?

enragednuke:
I thought I posted a version of the UI it has now, but I guess I forgot. Right now there's nothing limiting the data (it's using a similar structure to how it's structured originally, but it removes the list-formatting it used and makes it smoother looking using divs and CSS). I'm still looking into how I can condense it, but given that there may be really long names down the line, it's looking like the way I have it now is the best approach.

dfaulken:
Sounds good. I'd love to see a screenshot, but then this should be good.

enragednuke:
No clue what I did, but submitting an application (issue on [slack](http://transit-programmers.slack.com)) now magically works.

Here's the redesigned UI

screen shot 2016-02-10 at 4 23 43 pm

dfaulken:
You predicted merge conflicts, and we seem to have some here, but this is definitely good to merge in once those are resolved. Nice work, and sorry this PR fell under the radar between bojangles and aggressive-epsilon starting up.

enragednuke:
It also fell under the radar because I had to go back and fix my other PRs that I left during downtime. This isn't done yet though, I still need to clone the design to the other sections (and of course solve the merge conflicts) *then* it'll be good-to-go.

enragednuke:
Since there are so many merge conflicts at this point, I'm gonna recreate the branch and rewrite this stuff and push it in a new PR.

dfaulken:
:+1:

PR #202, opened 2016-02-05
Added Long-Text Response Type

### LINKED ISSUE #187: Add long-text data type to Question
... with proper handling in the view.

It'd be nice if we could ask ~~more extended questions~~ questions which require a more extended answer.

akaplo:
Closes #187.  Updated the Question model to include a "long-text" response type.  Added logic to handle this new type in drafts/show and templates/show.  

Also fixed a routing issue in student_dashboard that was returning an error whenever "Submit application for {name}" was being clicked on.

screen shot 2016-02-05 at 12 51 50 pm


dfaulken:
I wouldn't mind that text area being larger - say filling to the width of its bounding box and being 4 or 5 rows tall.

akaplo:
How does this look?

screen shot 2016-02-05 at 1 28 34 pm

dfaulken:
Much better!

PR #206, opened 2016-02-16
Routing errors

### LINKED ISSUE #203: Snakecase application template routes
The routing we implemented in #193 uses position and department names as is, which is a little awkward when spaces are involved, eg. `/applications/Special%20Transportation/Van%20Driver`, which is the generated URL for the ST application.

It'd be nice to match them by snakecase. The good news is that the bus application currently works fine - since we did implement case-insensitive matching and 'Bus' and 'Operator' have no spaces.

Anbranin:
closes #203 

Applications now viewable by a slug. It's not snakecased, but with dashes in between the words, which is different but arguably better because with dashes one does not have to press the shift key, at least on the majority of keyboards.

Some integration tests were written and a couple view tests were replaced, more will be written presently.


spec/features/application_draft_edit_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+  let!(:bottom_question) do
+    create :question, number: 3,
+                      application_draft: draft
+  end
+  before :each do
+    when_current_user_is :staff, integration: true
+    visit edit_draft_url(draft)
+  end
+  context 'moving questions' do
+    it 'contains button to move middle question up' do
```
akaplo: Isn't this more of a view spec than an integration spec?  I picture view specs as being "this button exists and this field exists" while integration tests say "clicking this button does this thing."  Am I incorrect in this?

app/controllers/application_templates_controller.rb
```diff
@@ -14,14 +14,18 @@ def show
 
   def toggle_active
     @template.toggle! :active
+    if @template.active
+      show_message :active_application,
+                   default: 'The application is now active'
```
dfaulken: Does this follow the same format as the other default messages?

config/application.yml.example
```diff
@@ -33,6 +34,10 @@ on_application_denial:
 # Unless otherwise marked, these messages are shown only to staff members.
 
 messages:
+  active_application: 'The application is now active'
+    # message to display if the application has been activated
```
dfaulken: These comments don't follow the same format as the ones that are there.

spec/features/application_draft_edit_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+    end
+  end
+  it 'has inputs to edit the question prompts' do
+    (0...draft.questions.count).each do |index|
+      expect(page.has_field?("draft[questions_attributes][#{index}][prompt]",
+                             type: 'textarea'))
+    end
+  end # of moving questions context
+  it 'allows selection of the data type of each question' do
+    all('select').each do |field|
```
dfaulken: What if we put other select tags on the page?

spec/features/application_draft_edit_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+      expect(field[:name]).to include 'data_type'
+    end
+  end
+  it 'has a checkbox to determine whether the question is required' do
+    all('checkbox').each do |field|
```
dfaulken: What if we put other checkboxes on the page?

spec/features/toggle_active_application_template_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    end
+  end
+  context 'on editing application drafts page' do
+    before :each do
+      visit edit_draft_path(draft)
+    end
+    context 'application is inactive' do
+      let(:application) { create :application_template, active: false }
+      it 'has a button to activate the application' do
+        expect(page).to have_button 'Activate application'
```
dfaulken: As we noted in person, it's fine to remove this line.

spec/features/application_draft_edit_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+  it 'has a button to bring the user to a preview draft page' do
+    click_button('Preview changes')
+    expect(page.current_url).to eql draft_url(draft)
+  end
+  context 'clicking save changes and continue editing button' do
+    it 'will add new questions to the draft' do
+      # draft starts with 3 questions
+      expect(draft.questions.count).to be 3
+      # fill in the prompt of the new question
+      fill_in(all('textarea').last[:name], with: 'Stuff')
```
dfaulken: I'm not sure what this line does.

dfaulken:
I'm going to merge this so that @akaplo and @bgregg can use your routing code, and all four of us can also look at my suggestions for your integration tests.

PR #207, opened 2016-02-19
Integration Test for filling out an application

bgregg:


spec/features/filling_out_application_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    visit application_path(application)
+  end
+
+  context 'answered all required questions' do
+    it 'redirects to the student dashboard' do
+      click_on 'Submit application'
+      expect(page.current_url).to eql student_dashboard_url
+    end
+
+    it 'renders a flash message' do
```
dfaulken: Very small thing, but since we're expecting a specific message, it'd be nice for our test description to reflect that.

spec/features/filling_out_application_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+      expect_flash_message(:application_receipt)
+      click_on('Submit application')
```
dfaulken: Not something I think needs fixing, but it's worth noting that you can leave both sets of parentheses out of this test:

```ruby
expect_flash_message :application_receipt
click_on 'Submit application'
```

It's a matter of personal preference - mine is to write as few non-alphanumeric characters as possible in my Ruby code.

dfaulken: Sorry, I meant to include the opposing viewpoint - a lot of people find parentheses a perfectly natural way of indicating the logical grouping of expressions, like in math, which is an even more widespread language than Ruby.

It's a good idea for readability's sake to keep the major points of code style consistent across a codebase, but this is such a minor point, that again, I don't think you need to go back and do anything about it. I just wanted to be informative. :smile:

dfaulken:
One very minor test description edit and this is good to go. Nice work.

dfaulken:
And :+1: for an atomic PR!

PR #212, opened 2016-02-24
Configurable EEO Requests

### LINKED ISSUE #189: Enable/disable EEO data per ApplicationTemplate

akaplo:
Closes #189.  Includes functionality similar to #133:

- When editing Drafts, adds a button to enable or disable EEO questions on the application

- When previewing a Draft, conditionally displays EEO questions

- When viewing a Template, conditionally displays EEO questions.

Added a migration for a new column in `ApplicationTemplate`: a `boolean` called `eeo_enabled`, which defaults to `true`.

Added a controller action `toggle_eeo_enabled` to `application_templates_controller`.  Controller test forthcoming.

One question before this gets merged (perhaps for @Anbranin): in the HAML, I used the same tags as you did in #189 to display the buttons, but they look VASTLY different (in terms of style).  Any idea why?

### DIFF
--- a/app/controllers/application_templates_controller.rb
+++ b/app/controllers/application_templates_controller.rb
@@ -22,6 +22,16 @@ class ApplicationTemplatesController < ApplicationController
     end
     redirect_to :back
   end
+  
+  def toggle_eeo_enabled
+    @template.toggle! :eeo_enabled
+    msg = 'You have enabled EEO data requests on your application.'
+    if !@template.eeo_enabled 
+      msg = 'You have disabled EEO data requests on your application.'
+    end
+    show_message :active_application, default: msg
+    redirect_to :back
+  end
 
   private
 
--- a/app/views/application_drafts/show.haml
+++ b/app/views/application_drafts/show.haml
@@ -47,23 +47,24 @@
                 required: question.required?
             - if question.required?
               %span.ast *
-      .heading Voluntary Equal Employment Opportunity Information
-      .explanation
-        The University of Massachusetts and the Pioneer Valley
-        Transit Authority are equal opportunity employers. To help
-        us ensure that we are complying with EEO policies, please supply
-        the information requested below.
-      .field
-        .label
-          Please select the ethnic group with which you most closely identify:
-        = select_tag :ethnicity,
-          options_for_select(ApplicationRecord::ETHNICITY_OPTIONS),
-          include_blank: 'Prefer not to answer'
-      .field
-        .label Please select the gender with which you most closely identify:
-        = select_tag :gender,
-          options_for_select(ApplicationRecord::GENDER_OPTIONS),
-          include_blank: 'Prefer not to answer'
+      - if @draft.application_template.eeo_enabled
+        .heading Voluntary Equal Employment Opportunity Information
+        .explanation
+          The University of Massachusetts and the Pioneer Valley
+          Transit Authority are equal opportunity employers. To help
+          us ensure that we are complying with EEO policies, please supply
+          the information requested below.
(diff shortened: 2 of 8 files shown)


app/controllers/application_templates_controller.rb
```diff
@@ -23,6 +23,20 @@ def toggle_active
     redirect_to :back
   end
 
+  def toggle_eeo_enabled
+    @template.toggle! :eeo_enabled
+    if @template.eeo_enabled
+      show_message :eeo_enabled,
+                   default:
+      'You have enabled EEO data requests on your application.'
```
dfaulken: I think you could probably pare this down to something shorter (and more meaningful) like

```
EEO data will [not] be requested for this application.
```

which would mean you can avoid splitting a key/value pair with a newline.

app/views/application_drafts/show.haml
```diff
@@ -47,23 +47,24 @@
...
-          Please select the ethnic group with which you most closely identify:
-        = select_tag :ethnicity,
-          options_for_select(ApplicationRecord::ETHNICITY_OPTIONS),
-          include_blank: 'Prefer not to answer'
-      .field
-        .label Please select the gender with which you most closely identify:
-        = select_tag :gender,
-          options_for_select(ApplicationRecord::GENDER_OPTIONS),
-          include_blank: 'Prefer not to answer'
+      - if @draft.application_template.eeo_enabled
```
dfaulken: Better to use the predicate method here (`#eeo_enabled?`).

app/views/application_templates/show.haml
```diff
@@ -66,21 +66,22 @@
...
-        Please select the ethnic group with which you most closely identify:
-      = select_tag :ethnicity,
-        options_for_select(ApplicationRecord::ETHNICITY_OPTIONS),
-        include_blank: 'Prefer not to answer'
-    .field
-      .label Please select the gender with which you most closely identify:
-      = select_tag :gender,
-        options_for_select(ApplicationRecord::GENDER_OPTIONS),
-        include_blank: 'Prefer not to answer'
+    - if @template.eeo_enabled
```
dfaulken: Predicate method here also.

spec/controllers/application_templates_controller_spec.rb
```diff
@@ -112,4 +112,60 @@
       end
     end
   end
+  describe 'POST #toggle_eeo_enabled' do
+    before :each do
+      department = create :department, name: 'Bus'
+      position = create :position, department: department, name: 'Operator'
+      @template = create :application_template, position: position
+    end # end before each
```
dfaulken: In general, it shouldn't be needed to comment on end tags in Ruby - better to just keep blocks nice and short.

The only exceptions in this context might be the top-level `describe` block - nothing else here should span more than a few lines at a time.

dfaulken:
The CC failure is okay - I agree there's some duplication there, but the functionality is separate, and I don't see a nice way of abstracting common logic.

Could we see some UI screenshots?

Looks good other than my nitpicks - very thorough.

app/controllers/application_templates_controller.rb
akaplo: Fixed.

akaplo:
Thanks!  I think I fixed everything.

Some screenshots below:

Setting EEO with a button:
screen shot 2016-03-01 at 11 48 35 am

Previewing a draft when EEO has been disabled:
screen shot 2016-03-01 at 11 48 42 am

A student's view of an application with EEO disabled:
screen shot 2016-03-01 at 11 51 43 am

PR #219, opened 2016-03-24
Phone number formatting

### LINKED ISSUE #204: Phone number auto-fomatting?
Couldn't help but notice that Don added to the phone number fields, e.g.:

screen shot 2016-02-07 at 9 13 10 pm

Perhaps we can help (with Javascript)?

Anbranin:
Phone number auto-formats on draft show page and application template show page. The JS will latch on to anything with the class ".phone".

New question type of "phone number" can be selected from question type dropdown.

Per @enragednuke's suggestion, I used a gem, specifically a rails-asset.

This pull request is ready to be merged whenever ready.

closes #204


dfaulken:
Some Codeclimate issues here- not entirely sure what happened with Travis but I imagine it'll resolve if you push to fix the CC issues

app/assets/javascripts/phone_number_formatting.js.coffee
```diff
@@ -0,0 +1,37 @@
...
+      if phone.val().length == 4
+        phone.val phone.val() + ')'
+      if phone.val().length == 5
+        phone.val phone.val() + ' '
+      if phone.val().length == 9
+        phone.val phone.val() + '-'
+
+      # Allow numeric, tab, backspace, delete keys only
+      key == 8 or key == 9 or key == 46 or key >= 48 and
+      key <= 57 or key >= 96 and key <= 185
```
enragednuke: What do these two lines do? They look like a boolean expression in the middle of nowhere (I assume it's either coffee or some weird interaction I don't know?)

enragednuke: Actually nevermind, probably an implicit return statement, didn't realize it was the end of the function

enragednuke:
So I was looking at the JS you based the formatter on and I found a bug (not sure if it applies to the code on your local branch since I'm not sure how to grab branches from other peoples' forks)

If you completely erase the string, it will remove the initial '(' as well. It will continue the other formatting, so you could end up with strings like this: `$224) 123-1423`

enragednuke:
In light of this, it may be worth looking into a gem version of the jQuery Masked Input, seen in action [here](http://jsfiddle.net/SO_AMK/SEXAj/). [Source for the plugin here](https://cloud.github.com/downloads/digitalBush/jquery.maskedinput/jquery.maskedinput-1.3.js)

Anbranin:
Thanks @enragednuke!! I'll look into this.

PR #223, opened 2016-03-28
application shows for staff whether active or not. feature tested.

### LINKED ISSUE #222: Application template not viewable by staff when inactive
The application template currently does not display at all when it has been marked as inactive. It should be viewable by staff at all times, just not by others. Additional if/then logic is needed.

Anbranin:
closes #222

### DIFF
--- a/app/views/application_templates/show.haml
+++ b/app/views/application_templates/show.haml
@@ -14,7 +14,7 @@
       = button_to 'Activate application',
           toggle_active_application_url(@template)
 
-- if (@current_user.try :staff?) || @template.active?
+- if @current_user.try(:staff?) || @template.active?
   .form
     %h1.title Application for #{@template.position.name_and_department}
     %strong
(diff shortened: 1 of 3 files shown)


app/views/application_templates/show.haml
```diff
@@ -14,7 +14,7 @@
       = button_to 'Activate application',
           toggle_active_application_url(@template)
 
-- if @template.active?
+- if (@current_user.try :staff?) || @template.active?
```
dfaulken: More idiomatically

```ruby
@current_user.try(:staff?) || @template.active?
```

spec/features/application_record_creation_spec.rb
```diff
@@ -21,6 +18,15 @@
       expect(find_field('Last name').value).to eql student.last_name
       expect(find_field('Email').value).to eql student.email
     end
+    context 'application template has been marked as inactive' do
+      before :each do
+        application_template.update active: false
+      end
+      it 'shows text explaining that the application is unavailable' do
+        expect page.has_text?
```
dfaulken: `expect(page).to have_text ...` would read better

PR #225, opened 2016-03-29
Bootstrap Integration (Staff Dashboard View & Application Template Show)

enragednuke:
Recreation of the #199 PR with no merge conflicts (it was very far behind and with the dumb 'database wiping' incident over spring break, it was easier just to recreate the branch in its entirety and just push this).

Still need to move the CSS to the other fields on the page, will comment again when that is done. Feel free to scan the code until then to make sure I didn't accidentally delete something wrongly during merge conflicts.

*Disclaimer*: There will likely be a side-effect of making other forms look very whacky until they're suited up with Bootstrap elements. @dfaulken said not to try and modify every page in one gigantic PR, so here's the first tidbit of it. Once this is merged, I'll work on getting the other views set up (so the addition of bootstrap doesn't appear in every PR).

PR #226, opened 2016-03-30
Bootstrap integration (Staff Dashboard & Application Template show)

enragednuke:
As per request by @dfaulken, this is now pushing to a separate `refactor` branch like `ft-refactor` on `round-three`.

See the old PR for visuals #225

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,6 @@
 source 'https://rubygems.org'
 
+gem 'bundler', '>= 1.8.4'
 gem 'coffee-rails'
 gem 'factory_girl_rails'
 gem 'haml'
@@ -20,7 +21,13 @@ gem 'uglifier'
 source 'https://rails-assets.org' do
   gem 'rails-assets-wenzhixin--multiple-select'
   gem 'rails-assets-jquery.maskedinput'
+  gem 'rails-assets-bootstrap'
 end
+
+group :assets do
+  gem 'twitter-bootstrap-rails'
+end
+
 group :production do
   gem 'exception_notification'
 end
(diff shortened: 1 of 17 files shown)


Gemfile
```diff
@@ -1,5 +1,6 @@
 source 'https://rubygems.org'
 
+gem 'bundler', '>= 1.8.4'
```
dfaulken: Does it work for bundler to bundle bundler?

app/assets/stylesheets/application_drafts/edit.scss
```diff
@@ -23,13 +23,14 @@ html {
...
   display: flex;
+  padding-bottom: 20px;
   .form_column{
     flex:3;
     //list-style-type:none;
     .header_row{
       padding: .5em;
       display: flex;
-      height: 2em;
+      //height: 2em;
```
dfaulken: Better to just remove

dfaulken:
Looks good; thanks for bearing with me on this process.

Other than the bundler question this looks good.

Gemfile
werebus: Actually, I don't see why it wouldn't.  But it's also probably unnecessary.

enragednuke: I likely put it there for *some* reason in the original PR, but it probably doesn't matter now. I'll see when I get in.

app/assets/stylesheets/application_drafts/edit.scss
enragednuke: I agree, I think I was messing with various values and accidentally left it in. Must've missed this when I was scanning for comments.

enragednuke:
@dfaulken The bundler gem wasn't needed (I'm still trying to think of why it was there in the first place). As long as checks pass it should be good to go.

PR #227, opened 2016-04-04
Add the ability to reload data from older applications (of the same type)

### LINKED ISSUE #28: Filling in old applications
When a student resubmits an application for a position, they should have the option (which configuration will enable or disable) to fill the fields based on their previous application. It will go by question name.

enragednuke:
Closes #28 

Some of this stuff can probably be tested, so it's probably not 1000% finished yet (nor have I run the local tests, just wanted to push this before I left for the day for review, so I assume checks will fail)

In the dump-y non-bootstrap interface, here's the look:

screen shot 2016-04-04 at 6 50 35 pm

General idea:

 1. When saving a AppRecord, it will now save the question's ID in the data-arrays
 2. When you open a AppTemplate (to apply), it will now display an option to fill in the fields with data from past applications (to the same position, if they exist)
 3. Maps the 4-element sub-arrays into 2-element sub-arrays which are then converted into a hash using `.to_h`
 4. Using `question.id` as a key, it will attempt to rip previous answers from the hash. If it fails, it'll default to an empty field

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -9,13 +9,20 @@ module ApplicationHelper
   def parse_application_data(data)
     questions = []
     data.each do |k, v|
-      match_data = k.match(/^(prompt|response|data_type)_(\d+)$/)
+      match_data = k.match(/^(prompt|response|data_type)?_?(\d+)$/)
       next unless match_data.present?
-      input_type, number = match_data.captures
-      questions[number.to_i] ||= []
-      input_types = %w(prompt response data_type)
-      index = input_types.index input_type
-      questions[number.to_i][index] = v
+      captures = match_data.captures.select{|c| !c.nil? }
+      if captures.length == 2
+        input_type, number = match_data.captures
+        questions[number.to_i] ||= []
+        input_types = %w(prompt response data_type)
+        index = input_types.index input_type
+        questions[number.to_i][index] = v
+      elsif captures.length == 1
+        q_id, number = v.split('-')
+        questions[number.to_i] ||= []
+        questions[number.to_i][3] = q_id.to_i + 1
+      end
     end
     questions
   end
@@ -28,6 +35,14 @@ module ApplicationHelper
 
   def should_show_denied_applications?
     configured_value [:on_application_denial, :notify_applicant], default: true
+  end
+
+  def should_allow_resubmission?
+    configured_value [:on_application_denial, :allow_resubmission], default: true
+  end
+
+  def should_allow_form_refilling?
+    configured_value [:on_application_denial, :fill_form_with_old], default: true
   end
 
   def text(name)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -9,20 +9,24 @@ module ApplicationHelper
   def parse_application_data(data)
     questions = []
     data.each do |k, v|
-      match_data = k.match(/^(prompt|response|data_type)?_?(\d+)$/)
+      match_data = k.match(/^(prompt|response|data_type)?_?(\d+)$/) ||
+                   v.match(/^(\d+)\-(\d+)$/)
       next unless match_data.present?
-      captures = match_data.captures.select{|c| !c.nil? }
-      if captures.length == 2
-        input_type, number = match_data.captures
-        questions[number.to_i] ||= []
-        input_types = %w(prompt response data_type)
-        index = input_types.index input_type
-        questions[number.to_i][index] = v
-      elsif captures.length == 1
-        q_id, number = v.split('-')
-        questions[number.to_i] ||= []
-        questions[number.to_i][3] = q_id.to_i + 1
-      end
+
+      input_type, number = match_data.captures
+      questions[number.to_i] ||= []
+      input_types = %w(prompt response data_type)
+
+      # Because we cannot match "all numeric values" for input_type, we
+      # must use a || 3 since if you .index(\d+) it will return nil here
+      # which the || 3 will catch.
+      index = input_types.index(input_type) || 3
+      questions[number.to_i][index] =
+        if index == 3 # a.k.a. If it is a numeric number (Q_ID)
+          input_type.to_i + 1
+        else
+          v
+        end
     end
     questions
   end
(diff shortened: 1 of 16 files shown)


app/models/user.rb
```diff
@@ -30,4 +30,8 @@ def proper_name
   def student?
     !staff?
   end
+
+  def old_applications(position_id)
```
dfaulken: It'd be more idiomatic to pass a whole Position object.

app/helpers/application_helper.rb
```diff
@@ -9,13 +9,20 @@ def configured_organization_name
...
+      if captures.length == 2
+        input_type, number = match_data.captures
+        questions[number.to_i] ||= []
+        input_types = %w(prompt response data_type)
+        index = input_types.index input_type
+        questions[number.to_i][index] = v
+      elsif captures.length == 1
+        q_id, number = v.split('-')
+        questions[number.to_i] ||= []
+        questions[number.to_i][3] = q_id.to_i + 1
```
dfaulken: Yeah.... but again, ID doesn't really work. If questions are deleted then ID isn't a useful reference anymore.

app/models/application_record.rb
```diff
@@ -50,6 +50,10 @@ def add_response_data(prompt, response)
     update(data: data << [prompt, response])
     self
   end
+  
+  def questions_hash
+    data.map{|arr| [arr[3], arr[1]]}.select{|arr| [arr[0], arr[1]].all?}.to_h
```
dfaulken: I think this could be elongated for readability's sake.

app/views/application_templates/show.haml
```diff
@@ -14,13 +14,27 @@
...
+        %table.col-md-8.col-md-offset-2
+          %tr
+            %th
+              Submitted
+          - @old_applications.each do |app|
+            %tr
+              %td
+                = format_date_time app.created_at
+              %td.load_column
+                = link_to 'Load reponses', url_for(params.merge(:load_id => app.id))
```
dfaulken: Better to do it outright:

```ruby
link_to 'Load responses', application_template(@template, load_id: app.id)
```

lib/application_configuration.rb
```diff
@@ -7,7 +7,13 @@ module ApplicationConfiguration
...
   def configured_value(config_path, options = {})
     value = CONFIG.dig(*config_path)
-    if value.present? then value
+
+    # This is a peculiar situation because value will return {}
+    # on failed queries. However, it is entirely possible that
+    # a query will yield false (as a configured value), so we cannot
+    # use standard .present? or .blank?. We also cannot use .empty?
+    # because TrueClass and FalseClass do not support it
+    if value != {} then value
```
dfaulken: Good catc

app/helpers/application_helper.rb
enragednuke: I don't see the issue. If they're deleted, then it just won't read from the hash? If the question is deleted, then the data associated with said question is irrelevant as well. Simply associating based on question name is a poor idea because there can be different contexts outside of the question. For example:

**As a child,**

 * Life goal?
 * Favorite super hero?

**As you are now**

 * Life goal?
 * Favorite super hero?

There's no way to differentiate between the two besides their ID. They will have the same question prompt and data type. You can't go based off the 'context header' part because in our system those headers are **also** questions.

I don't see a solution that can work for every question other than this.

config/application.yml.example
```diff
@@ -35,7 +35,12 @@ on_application_denial:
   # For instance, you may wish to provide feedback on steps applicants
   # could take in order to become eligible for the position in the future.
   notify_of_reason: true
-
+  # Whether or not an applicant will be allowed to resubmit an application
+  # to a declined application
```
dfaulken: I'd go with "whether or not an applicant will be allowed to resubmit an application for a particular position when they have previously had an application for that position denied"

app/models/user.rb
enragednuke: I considered that, but wasn't really sure where was better to do the logic.

app/helpers/application_helper.rb
dfaulken: That's a good point! We're never going to come up with a perfect solution in a data retrieval situation like this anyway. The good thing about your solution is that it has a very good chance of working well if the past application was recent, and a lower chance of working well with less recent applications. You've convinced me.

dfaulken:
This is impressively fast work. Just a few comments on code style.

config/application.yml.example
enragednuke: :+1:

app/views/application_templates/show.haml
enragednuke: Is it possible to pass the template directly in since this is the page with the new `/department/position` routing?

app/models/application_record.rb
enragednuke: Elongated in the sense of tabbing it to multiple lines and commenting it or what? Not entirely sure how I'd extend the code itself.

app/views/application_templates/show.haml
dfaulken: Yep, friendly-id takes care of that.

app/models/application_record.rb
dfaulken: Something that makes it clearer what is in each position of the array. Intermediate variables.

enragednuke:
The rubocop and rspec (as well as stuff mentioned in comments) have been added/fixed (though GitHub seems to be slow right now so it hasn't recognized that I pushed the commits?). Will need to add my own tests for the new functionality and then it'll be good to go.

enragednuke:
@dfaulken Specs added, all checks pass.

100% coverage except for [THIS](https://github.com/umts/jobapps/pull/227/files#diff-93f65992721769152ca0e1d37ed738a8R17) line. No matter what I tried (it wasn't included in the commit), I couldn't figure out how in the world to test for it. I did what I thought was right (attempt to pass a `load_id` with the value of an existing record's ID, then test the instance variable) but it was no bueno.

enragednuke:
Got it. Turns out I needed the `when_current_user_is` line, otherwise it wouldn't render the value properly (which makes sense in retrospect).

All good if checks pass.

dfaulken:
Merge conflicts ⚠️

enragednuke:
Fixed @dfaulken

PR #233, opened 2016-04-15
EEO page displays number of people from each eeo group that was interviewed

### LINKED ISSUE #229: Request: Display interview information with EEO data
Don requested that the number of a given set (EEO data-set) of people that got interviews be displayed on the EEO data page.

Anbranin:
This is what Don asked for, and it's done.

screen shot 2016-04-15 at 4 44 42 pm

closes #229


spec/models/application_record_spec.rb
```diff
@@ -104,6 +104,21 @@
     end
   end
 
+  describe 'interview_count' do
+    before :each do
+      create :application_record, id: 1
```
dfaulken: Shouldn't create with IDs

dfaulken:
Good; just a test restructure needed - tests shouldn't depend on specific IDs of things.

app/models/application_record.rb
```diff
@@ -68,27 +72,33 @@ def self.gender_eeo_data(start_date, end_date, department_ids)
                                                   .in_department(department_ids)
     all_genders = GENDER_OPTIONS | gender_records.pluck(:gender)
     all_genders.map do |option|
-      [option, gender_records.where(gender: option).count]
+      specific_records = gender_records.where(gender: option)
+      [option, specific_records.count, specific_records.interview_count]
     end
   end
 
+  # rubocop:disable Metrics/AbcSize
```
enragednuke: That's a nifty line. Does this disable the Rubocop check for "too much stuff/logic in one method"?

Anbranin: YUP. I like that method how it is, Rubocop can shut up.
On that note, I'm changing the method... because it's not dry enough. :(

dfaulken: ... which should decrease the abc size also. Be interesting to see if this passes Rubocop check once you've done that.

spec/models/application_record_spec.rb
```diff
@@ -106,10 +106,10 @@
...
   describe 'interview_count' do
     before :each do
-      create :application_record, id: 1
-      create :application_record, id: 2
-      create :interview, application_record_id: 1
-      @collection = ApplicationRecord.where(id: 1)
+      record_1 = create :application_record
+      record_2 = create :application_record
+      create :interview, application_record: record_1
+      @collection = ApplicationRecord.where(id: record_1.id)
```
enragednuke: @dfaulken Why not just make `record_1` an instance variance instead? This seems like an unnecessary database query.

Even if it was to be kept, wouldn't it be better to replace it with `.find(record_1.id)` or add `.take` to the end to account for the situation when the query returns more than one result? (abstraction for future addition to test suite since `.where` returns an ActiveRecord relation)

spec/models/application_record_spec.rb
```diff
@@ -104,6 +104,21 @@
     end
   end
 
+  describe 'interview_count' do
+    before :each do
+      record_1 = create :application_record
+      record_2 = create :application_record
+      create :interview, application_record: record_1
+      @collection = ApplicationRecord.where(id: record_1.id)
```
enragednuke: (put this here because Github called my other post on an outdated diff)

Is there a reason to not just make record_1 an instance variance instead? This seems like an unnecessary database query.

Even if it was to be kept, wouldn't it be better to replace it with .find(record_1.id) or add .take to the end to account for the situation when the query returns more than one result? (abstraction for future addition to test suite since .where returns an ActiveRecord relation)

dfaulken: @Anbranin's code is correct - this is testing an ActiveRecord scope, so it needs to be called on an ActiveRecord relation.

One application record has an interview, one does not - the scope should find that only one does and return the number 1.

dfaulken: Actually, now that you mention it, the collection should contain both:

```ruby
@collection = ApplicationRecord.where(id: [record_1.id, record_2.id])
```

enragednuke: Isn't calling `.interview_count` calling on a ApplicationRecord object? I'm confused as to how that's an AR relation.

If, for whatever reason, the `.where` yielded more than one object (it shouldn't though, because it's on an ID) then this code would break because there would be more than one object in the AR relation.

I still don't see why using `.find` or utilizing `record_1` as an instance variance *wouldn't* be better here.

enragednuke: Oh, that variable was meant to include **both** of them. That makes much more sense (I saw your comment after I posted my last comment)

enragednuke: But then wouldn't this code break? I thought calling a model method on an ActiveRecord relation that contains more than one object will throw an exception because it's like trying to call `["1", "2"].to_i`.

dfaulken: An ActiveRecord scope is a method defined on ActiveRecord relations of a particular class. The code is correct.

http://guides.rubyonrails.org/active_record_querying.html#scopes

dfaulken: In particular, if it were broken, the Travis build would have failed.

enragednuke: You're right. I totally forgot that `interview_count` was a scope. For some reason I was interpreting it as a model method (even though I read it lol). Derp.

Anbranin:
kill me

app/models/application_record.rb
```diff
@@ -88,15 +88,18 @@ def self.eeo_data(start_date, end_date, department_ids)
...
-    end
-    records[:female_ethnicities] = all_ethnicities.map do |ethnicity|
-      specific_records = ethnicity_records
-                         .where ethnicity: ethnicity, gender: 'Female'
-      [ethnicity, specific_records.count, specific_records.interview_count]
+    all_ethnicities.map do |ethnicity|
+      combined_records = ethnicity_records.with_gender
+                                          .where(ethnicity: ethnicity)
+      male_records = combined_records.where ethnicity: ethnicity, gender: 'Male'
+      records[:male_ethnicities] << [ethnicity,
```
dfaulken: So, this is still the same code repeated twice with just 'Male' and 'Female' being different. Let's work together on this tomorrow.

Anbranin:
I can't even.
😭 
screen shot 2016-04-21 at 7 23 55 pm

lib/application_configuration.rb
```diff
@@ -13,7 +13,7 @@ def configured_value(config_path, options = {})
     # a query will yield false (as a configured value), so we cannot
     # use standard .present? or .blank?. We also cannot use .empty?
     # because TrueClass and FalseClass do not support it
-    if value != {} then value
+    if !value.nil? then value
```
Anbranin: Dig, in fact, returns 'nil' when a missing key is hit:
http://ruby-doc.org/core-2.3.0/Hash.html#method-i-dig

app/models/application_record.rb
```diff
@@ -77,32 +77,47 @@ def pending?
...
+    sub_hash = {}
+    all_ethnicities = ETHNICITY_OPTIONS | combined_records.pluck(:ethnicity)
+    all_genders = GENDER_OPTIONS | combined_records.pluck(:gender)
+    all_genders.map do |gender|
+      sub_array = []
+      all_ethnicities.map do |ethnicity|
+        records = combined_records.where(ethnicity: ethnicity,
+                                         gender: gender)
+        sub_array << [ethnicity, records.count, records.interview_count]
+        sub_hash.store(:"#{gender}", sub_array)
```
dfaulken: I think

```ruby
sub_hash[gender] = sub_array
```

app/views/application_records/_combined_totals.haml
```diff
@@ -1,19 +1,14 @@
...
-      %th Total Applicants
-  - @records[:female_ethnicities].each do |ethnicity, count|
-    %tr
-      %td= ethnicity
-      %td= count
+- @records[:combined_data].each do |key, value|
+  %table#eeo_table
+    %thead
+      %tr
+        %th= "#{key}s by Ethnicity"
```
dfaulken: "Males" and "females" are OK, but this might end up looking not great if we incorporate more gender options.

app/models/application_record.rb
enragednuke: I think she wants it to be a symbol (note: this won't work if gender is an integer, but I don't think it is)

```ruby
sub_hash[gender.to_sym] = sub_array
``

File: lib/application_configuration.rb (hunk 6 above)
Comment (enragednuke) [inline lib/application_configuration.rb]: It definitely did not whenever I was writing this I'm about 95% sure of this. I remember something rendering on the page with the value `{}` because it passed this check.

File: lib/application_configuration.rb [hunk 9]
```diff
@@ -7,13 +7,8 @@ module ApplicationConfiguration
...
     value = CONFIG.dig(*config_path)
-
-    # This is a peculiar situation because value will return {}
-    # on failed queries. However, it is entirely possible that
-    # a query will yield false (as a configured value), so we cannot
-    # use standard .present? or .blank?. We also cannot use .empty?
-    # because TrueClass and FalseClass do not support it
-    if value != {} then value
+    # dig returns nil when key not found
+    if value.present? then value
```
dfaulken: @enragednuke's comment about `false` was correct, though, so in this case we'll do the C idiom:

```ruby
if value then value
```

app/models/application_record.rb
dfaulken: There's no reason to make it a symbol.

lib/application_configuration.rb
dfaulken: Not sure how that happened, but the Ruby docs do back `nil` up.

lib/application_configuration.rb
Anbranin: Oh!

lib/application_configuration.rb
```diff
@@ -8,7 +8,9 @@ module ApplicationConfiguration
   def configured_value(config_path, options = {})
     value = CONFIG.dig(*config_path)
     # dig returns nil when key not found
-    if value.present? then value
+    # but, it is possible that a query will yeild false
```
dfaulken: yield

dfaulken:
Looks good; just merge conflicts.

PR #234, opened 2016-04-20
Added EEO Data page spec

### LINKED ISSUE #232: Eeo data display page remains untested
It'd be nice to write some tests to make sure the data is displayed correctly. Maybe some integration tests vaguely modeled after the viewing_past_applications_table_spec.

enragednuke:
Closes #232 

Also, as a rule of thumb: CSS IDs should not be duplicated (much like IDs in Rails, which **can't** be duplicated). If you find a situation where you think something should have the same ID, it's likely supposed to be a class instead. Use IDs as a delimiter between objects of the same class or if there is only one object that would reasonably have said ID (like the `data-table`, but even then it seems more like a class than an ID). I've modified the CSS to reflect this as well.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -15,7 +15,7 @@ RSpec.configure do |config|
   end
   config.include FactoryGirl::Syntax::Methods
   config.include UmtsCustomMatchers
-  Dir['./spec/support/**/*_shared_example.rb'].sort.each { |f| require f }
+  Dir['./spec/support/**/*_shared_example.rb'].each { |f| require f }
 end
 
 # controller spec helper methods
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -15,6 +15,7 @@ RSpec.configure do |config|
   end
   config.include FactoryGirl::Syntax::Methods
   config.include UmtsCustomMatchers
+  Dir['./spec/support/**/*_shared_example.rb'].sort.each { |f| require f }
 end
 
 # controller spec helper methods
--- a/spec/support/shared_examples/data_page_shared_example.rb
+++ b/spec/support/shared_examples/data_page_shared_example.rb
@@ -1,0 +1,7 @@
+shared_examples 'a data page' do |tables|
+  it 'contains all requires tables' do
+    tables.each do |table_name|
+      expect page.has_table? table_name
+    end
+  end
+end
--- a/spec/support/shared_examples/data_page_shared_example.rb
+++ b/spec/support/shared_examples/data_page_shared_example.rb
@@ -1,7 +1,7 @@
-shared_examples 'a data page' do |tables|
+shared_examples 'a data page' do |table_ids:|
   it 'contains all requires tables' do
-    tables.each do |table_name|
-      expect page.has_table? table_name
+    table_ids.each do |table_id|
+      expect(page).to have_table table_id
     end
   end
 end
--- a/app/assets/stylesheets/application_records/eeo_display.scss
+++ b/app/assets/stylesheets/application_records/eeo_display.scss
@@ -1,4 +1,4 @@
-table#eeo_table{
+table.eeo_table{
   margin: 45px;
   border-collapse: collapse;
   text-align: left;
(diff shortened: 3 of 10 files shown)


enragednuke:
Now utilizes shared examples (which can be used on all pages like this and the past_applications page, aka pages that utilize tables and the like)

dfaulken:
There are plenty of HTML style guides out there that just flat-out say "don't use IDs."

spec/spec_helper.rb
```diff
@@ -15,6 +15,7 @@
   end
   config.include FactoryGirl::Syntax::Methods
   config.include UmtsCustomMatchers
+  Dir['./spec/support/**/*_shared_example.rb'].sort.each { |f| require f }
```
dfaulken: `sort`?

dfaulken:
I'm generally not a fan of shared examples, and I can go into that more if you want, but I think this is a fine way of using them.

For the argument, I might prefer keyword args to just an array, so that it's clear when you're invoking the example what that array stands for.

```ruby
it_behaves_like 'a data page', table_names: %w(a list of names)
```

spec/support/shared_examples/data_page_shared_example.rb
```diff
@@ -0,0 +1,7 @@
+shared_examples 'a data page' do |tables|
+  it 'contains all requires tables' do
+    tables.each do |table_name|
+      expect page.has_table? table_name
```
dfaulken: Better to do `expect(page).to have_table table_name`

enragednuke: I was just copying the spec page mentioned in the original issue. I'll swap this around.

spec/spec_helper.rb
enragednuke: This was a snippet I got from google (as the support directory is not natively supported). I didn't really question the sort much, but I imagine it wouldn't matter too much.

enragednuke:
I like IDs, but only as a delimiting attribute. I think flat out ignoring a major part of CSS is poor practice (in my opinion). It's also useful as a pseudo-comment in which it's a **id**entifier as to what that object is. It is also very useful for readability such that future employees can easily understand what it is without having to look into its content.

What's the reason you dislike shared examples? It's a convenient way of reducing the repetitiveness of specs that are similar. I can't really think how they'd be bad (but I'm sure you have a good reason for thinking so).

As for the key inputs, I'll try to have it work like that. Not sure if it's just as simple as giving it the same name as the loop variable in the shared example, but I'll see what I can do.

dfaulken:
Thoughtbot's [guide](https://robots.thoughtbot.com/ruby-2-keyword-arguments) to keyword arguments should be helpful to you.

As far as shared examples go, I'll write a blog post. (We're doing blogs today.)

enragednuke:
That's fun, I thought about trying something like that but I didn't think it'd actually work.

Are you publishing the blog post anywhere?

dfaulken:
Let's talk more about it when you're here this afternoon.

spec/support/shared_examples/data_page_shared_example.rb
enragednuke: Upon further inspection, the two aren't interchangeable like that it seems. Doing as you did yields failed tests (doesn't find `data_table` on either). I can't seem to figure out *how* they're different though. I thought it might work on the parent element (nope) or maybe I had to include the `#` to denote it's looking at ID (also nope).

enragednuke: So on the false positive thing, it turns out this snippet fails (says it returns false)

```ruby
expect(page.has_table?('DataTables_Table_0')).to be(true)
```

but this snippet passes

```ruby
expect page.has_table?('DataTables_Table_0')
```

enragednuke:
Should be patched up now.

spec/support/shared_examples/data_page_shared_example.rb
```diff
@@ -1,7 +1,7 @@
-shared_examples 'a data page' do |tables|
+shared_examples 'a data page' do |table_ids:|
```
dfaulken: 👍

PR #235, opened 2016-04-20
Added Dashboard link for students

enragednuke:
No GH issue with this, just an addition for an issue I found super annoying when writing other code.

Adds a 'Dashboard' link in the upper right if the user is a student much like the existing one for staff.
screen shot 2016-04-20 at 4 19 07 pm

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -58,14 +58,10 @@ module ApplicationHelper
   end
 
   def dashboard_path
-    # Just incase this method is called when a current user does not exist
-    # return the safe method of a link destination that does not move the
-    # page
-    return '#' unless @current_user
-
-    if @current_user.staff?
+    if @current_user.try :staff?
       staff_dashboard_path
     else
+      # We still want anonymous logins to redirect to student dashboard
       student_dashboard_path
     end
   end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -56,4 +56,17 @@ module ApplicationHelper
     text = SiteText.where(name: name).first.try :text
     render_markdown text if text.present?
   end
+
+  def dashboard_path
+    # Just incase this method is called when a current user does not exist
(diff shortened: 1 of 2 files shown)


enragednuke:
Alternative methods

```ruby
- dashboard_link =
- if @current_user.staff?
  - staff_dashboard_path
- else
  - student_dashboard_path

= link_to 'Logout', destroy_session_path
= link_to 'Dashboard', dashboard_link
```

```ruby
# But Dave doesn't like the ternary operator
= link_to 'Dashboard', @current_user.staff? ? staff_dashboard_path : student_dashboard_path
```

dfaulken:
I think there are two more straightforward solutions:

1. A helper method (rather than just some code) that figures out the correct path;
2. Just writing two links:

  ```haml
    - if @current_user.staff?
      = link_to 'Dashboard', staff_dashboard_path
    - else
      = link_to 'Dashboard', student_dashboard_path

dfaulken:
Also, you *can* omit the `?` for boolean columns, but you shouldn't. Think of it as just like any other predicate method.

enragednuke:
I like the helper method idea. Keeps the logic out of the view. Not sure how I didn't think of that.

app/helpers/application_helper.rb
```diff
@@ -56,4 +56,17 @@ def text(name)
     text = SiteText.where(name: name).first.try :text
     render_markdown text if text.present?
   end
+
+  def dashboard_path
+    # Just incase this method is called when a current user does not exist
```
dfaulken: I'd disagree, actually: there's a prominent use case for students using the site but not having created a user.

I would just change the condition to be `if @current.user.try :staff?`

enragednuke: I was actually split on using that since I didn't know how the links should work when there's the "anonymous" login (as seen in the developer login), was planning to ask about it today. Good to know, I'll swap this.

dfaulken:
Nice work.

PR #238, opened 2016-04-22
Fix dig expectations

Anbranin:
Dig returns "nil" when no key is found. Old method expected it to return {} and checked for that. Fixed.


werebus:
Looks good.

enragednuke:
@dfaulken @Anbranin 

I'd like to once again reiterate what I was talking about [here](https://github.com/umts/jobapps/pull/233/files/c7ca388bd3c33159df43e652dbe5e684fb1fb9eb#r60751715) and show you the following:

When I type the following into my console (feel free to try it yourself incase my console is bugged out): `configured_value(:a)` it returns `{}`. **Not** `nil` as we so expect. (Note: to use this in the console, type `include ApplicationHelper` or `include ApplicationConfiguration`)

So, as I mentioned, when you call this on an unknown dig path it will return an empty hash which **will pass** the check provided here and just return the empty hash.

I *did* warn this might occur, but apparently nobody tested it themselves before merging this.

/rant

Furthermore, I have not a single clue why it's doing this. But I can note that something funky is going on here. Note the following:

```ruby
2.3.0 :001 > a = {:a => 1, :b => 2}
 => {:a => 1, :b => 2}
2.3.0 :002 > a.dig(:c)
 => nil
2.3.0 :003 > a.dig(:b)
 => 2
```

As expected it, it works as we expect. However, in the `rails console` interface...

```ruby
2.3.0 :001 > CONFIG
 => (REALLY big hash, but one we expect)
2.3.0 :002 > CONFIG.dig(:organization_name)
 => "UMass Transit"
2.3.0 :003 > configured_value(:organization_name)
 => "UMass Transit"

# Now here comes the weird part
2.3.0 :004 > CONFIG.dig(:a)
 => {}
2.3.0 :005 > CONFIG
 => (REALLY big hash, HOWEVER the key :a => {} has been added!)
2.3.0 :006 > configured_value(:b)
 => {}
2.3.0 :007 > CONFIG
 => (REALLY big hash, HOWEVER the key :b => {} has been added as well!)
```

Perhaps has to do something with CONFIG being ripped from `application.yml`?

Issue #115, opened 2015-12-16
Confgurable reply_to on ApplicationTemplate


dfaulken:
Same reason as https://github.com/umts/ft-forms/issues/39.

Also I think we should come up with a better application denial email closing text than

> Best,
UMass Transit

enragednuke:
![screen shot 2016-04-04 at 1 36 03 pm](https://cloud.githubusercontent.com/assets/9220151/14257117/33165f9c-fa6a-11e5-8cf1-be35d0fd93c9.png)

wat

dfaulken:
lol

PR #239, opened 2016-04-22
Reply to

petermarathas:
Application templates now contain a configurable reply-to address.


app/mailers/jobapps_mailer.rb
```diff
@@ -8,22 +8,32 @@ class JobappsMailer < ActionMailer::Base
   def application_denial(application_record)
     @application_record = application_record
     @user = application_record.user
+    position = application_record.position
+    reply_to = position.application_template.present? ?
```
dfaulken: I agree with CodeClimate - better to avoid ternary statements

spec/factories/application_templates.rb
```diff
@@ -2,6 +2,7 @@
   factory :application_template do
     position
     active true
+    email 'steve@sharklazers.com'
```
dfaulken: Do application templates require an email to be valid?

dfaulken:
Does this close an issue?

spec/factories/application_templates.rb
petermarathas: Nope. Does that mean this should be removed?

dfaulken: Yes. Factories should only contain the attributes which are necessary in order to make a valid instance.

petermarathas:
Closes #115

petermarathas:
@dfaulken I added <a href="https://github.com/umts/jobapps/pull/239/files#diff-1717d3e3835cc67fc19f0620ff5bc59aR109">this</a> and <a href="https://github.com/umts/jobapps/pull/239/files#diff-fcfc74e9d1b177a47d76c4345065d388R39">these</a> methods to make it easier to get application template and position. I was getting annoyed by how much I had to write each time I wanted an interview's original template and the same for record. 

Also I added <a href="https://github.com/umts/jobapps/pull/239/files#diff-5ba8e59a70da381cbe96315ecc39389dR11">this check</a> because there were some tests in which a record's position wouldn't have a template. I'm not sure if there would ever be a real occurrence of an application record not having an application template.

petermarathas:
Nevermind, code climate didn't like those methods. Still though, would there ever be a time where an application record would exist but its position would not have an application template?

dfaulken:
Sure - if an application was made in the past and then the position was no longer offered, we might delete the template.

app/mailers/jobapps_mailer.rb
```diff
@@ -9,11 +9,11 @@ def application_denial(application_record)
     @application_record = application_record
     @user = application_record.user
     template = application_record.position.application_template
-    if template.present?
-      reply_to = template.email
-    else
-      reply_to = ''
-    end
+    reply_to = if template
```
dfaulken: Could also do

```ruby
reply_to = template.try(:email)
```

unless the mailer has a problem with a nil address rather than a blank string.

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 100.0
+    "covered_percent": 99.69
```
dfaulken: Oops - looks like both conditional branches aren't tested.

app/mailers/jobapps_mailer.rb
```diff
@@ -31,7 +36,12 @@ def interview_confirmation(interview)
   def interview_reschedule(interview)
     @interview = interview
     @user = interview.user
-    reply_to = interview.application_record.position.application_template.email
+    template = interview.application_record.position.application_template
+    reply_to = if template
```
dfaulken: Will the mailer accept a `nil` reply_to value? If so, you could just do

```ruby
reply_to = template.try :email
```

dfaulken:
I'll take care of finishing this PR up @petermarathas. Thanks for getting it this far

dfaulken:
Merging - I'm okay with those CodeClimate issues.

PR #257, opened 2016-06-23
Denial email

### LINKED ISSUE #224: Application denial email doesn't reference the position
This will pretty much be ameliorated by #115, but I noticed that when someone gets denied and they reply to the email, it's not 100% clear what position they were denied for.

### LINKED ISSUE #256: Remove route exception
Don't need [this particular exception][0] anymore.

[0]:  https://github.com/umts/jobapps/blob/1d80e9d258dcc860554085d06683936a3afe6b1e/config/routes.rb#L14

AkivaGreen:
I removed route exception and included application position in the denial email.
resolve #256 
resolve #224


dfaulken:
Good work.

PR #261, opened 2016-07-18
Pdf change

### LINKED ISSUE #260: PDF of Application Record doesn't look pretty enough
The Training Supervisor, who prints these application records, recently began having issues printing directly from the website. The PDF we gave him (with a 'print this page' button) was not pretty enough (his words), and printing the web page directly gave him a black sheet. What a waste of ink. I thought, then, that I would make the PDF better looking for him. These are the specific things he requested:

The explanation can not be displayed, that's fine
There should be padding between sentences, and padding horizontal, and vertical--basically everything needs to be more spacious.
Horizontal lines separating question/answer pairs
A box around the whole thing. Makes it nice and enclosed.
The email address of the applicant needs to be included in the header along with their name and the date they applied
Try removing the shading, maybe it'll be nicer without it

Update: He just told me he's now selecting 'save as PDF'  and the application record is printing EXACTLY like he wants, which is how it looks on the application_record/show page. Why even bother making a PDF, then?

I'd mark this 'low priority' if it was an available label.

Anbranin:
You guys I'm so good at making PDFs now.
I asked @sherson and @werebus about the size of the information in the header, they are both of different opinions. However, Donald Chapman says he likes it. I asked him if he'd stop printing the web page now and he said "probably" which is about as good as I'm gonna get I guess.

screen shot 2016-07-18 at 12 55 28 pm

closes #260


app/views/application_records/show.haml
```diff
@@ -1,15 +1,16 @@
-%h1 Application Record
+%h1.h2="#{@record.position.name} Application Record"
```
dfaulken: What's with all the h-tags that have a CSS class named after another, smaller h-tag?

dfaulken: And just as a reminder, if you want to string interpolate in HAML, you don't actually need quotes:

```haml
%h1.h2 #{record.position.name} Application Record
```

Anbranin: @sherson thought that the default size of the h1 was too big. I agreed.

app/pdfs/print_record_pdf.rb
```diff
@@ -5,41 +5,44 @@ class PrintRecordPdf < Prawn::Document
...
+        cells.padding = 12
         self.header = true
-        self.row_colors = %w(DDDDDD FFFFFF)
         self.column_widths = [column_width, column_width]
+        self.cell_style = { borders: [:bottom], border_width: 0.5 }
       end
     end
   end
 
   def data_rows
```
dfaulken: While we're here, this kind of data processing seems like it belongs in the model - that way we can test it, and it's not taking up space here.

app/views/application_records/show.haml
dfaulken: Everywhere? Or just here?

dfaulken: If it's everywhere, then a generic CSS rule on `h1` is in order.

If it's just here, then there two alternatives leap to mind which would communicate our intent a little better:

1. Wrap the page in some kind of div (`.smaller-headers`) and then write CSS specific to that:
  ```css
  .smaller-headers h1{ font-size: ??px; }
  ```
2. Writing a more descriptive CSS class to apply directly to the headings:
  ```css
  h1.we-want-it-to-be-a-little-smaller{ font-size: ??px; }
  ```

Anbranin: I pointed out all the other h1s on the page. He said, and I quote "In my opinion they _are_ too big, but I don't care."

app/pdfs/print_record_pdf.rb
Anbranin: Moved. Writing tests.

app/views/application_records/show.haml
Anbranin: Would that be assuming that people working on our app didn't know about
bootstrap and what those h classes are for?

dfaulken: Except that's not what those classes are for in Bootstrap - [they're about maintaining the font size while having a different display style](http://getbootstrap.com/css/#type-headings), not having an `h1` that is the same size as an `h2`.

dfaulken: As in, the idea is that you would normally use an `h1` tag, but if you want your text to be inline, you use `.h1`, but there's no reason to use both, and definitely no reason to use an `h1` tag and the `.h2` class, which is a nice trick, but doesn't really communicate what you want.

Anbranin: Ooooh okay!

Anbranin:
@sherson doesn't care enough about the size of the h1s in the app for me to bother changing them site-wide and then making sure all the other header tags are sized appropriately.

spec/models/application_record_spec.rb
```diff
@@ -67,6 +67,21 @@
     end
   end
 
+  describe 'data_rows' do
+    let :call do
+      @record.data_rows
+    end
+    before :each do
+      @record = create :application_record,
+                       data: [['a question', 'an answer']]
```
dfaulken: Isn't each element in `data` a 4-element array?

spec/models/application_record_spec.rb
```diff
@@ -67,6 +67,21 @@
...
+    end
+    it 'returns an array containing a header, then prompts and responses' do
+      expect(call).to eql [%w(Question Response), ['a question', 'an answer']]
+    end
```
dfaulken: Good start - how about the question filtering it does?

sherson:
I'm always happy to do more bikeshedding 😄

spec/models/application_record_spec.rb
Anbranin: Not at the time it was written, I suspect. Both add_data_types and add_response_data are only used in lib tasks--used when migrating data from the initial app state.

Anbranin: So here's a thought--should these tasks and methods be deleted, or merely updated?

dfaulken: >Not at the time it was written

Not at the time *what* was written? It looks like `data_rows` is expecting a 4-element array to be each element of data types. You just wrote these tests.

We can remove import tasks and related methods.

Anbranin: hahaha woops, I was looking at another test. Then I copied what was there to test mine without thinking. But cool yes I'll remove all that stuff.

PR #262, opened 2016-07-18
Subsidiary accounts can log in - Fix

### LINKED ISSUE #136: Subsidiary accounts can log in
Just accidentally logged into the site as @sherson with the "rtic" account.  It's not a major issue, but it does mean that the TRSpvs can log in as Don pretty easily since they both have access to an account that's "Don's"

In round-three we guard against this by checking whether 
```ruby
request.env["UMAPrimaryAccount"] == request.env["uid"]
```

See umts/round-three#29

AkivaGreen:
I want to resolve #136 with this pull request.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -9,7 +9,7 @@ class ApplicationController < ActionController::Base
   before_action :set_current_user
   before_action :redirect_unauthenticated
   before_action :access_control
-  before_action :primary_account
+  before_action :check_primary_account
   layout 'application'
 
   private
@@ -66,9 +66,11 @@ class ApplicationController < ActionController::Base
   end
   # rubocop:enable Style/AndOr
 
-  def primary_account
+  def check_primary_account
     if request.env['UMAPrimaryAccount'] != request.env['uid']
-      render file: 'sessions/unauthorized_subsidiary',
+      @UMAPrimaryAccount = request.env['UMAPrimaryAccount']
+      @uid = request.env['uid']
+      render file: 'sessions/unauthenticated_subsidiary',
              status: :unauthorized,
              layout: false
     end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -69,9 +69,8 @@ class ApplicationController < ActionController::Base
   def primary_account
     if request.env['UMAPrimaryAccount'] != request.env['uid']
       render file: 'sessions/unauthorized_subsidiary',
-            status: :unauthorized,
-            layout: false
+             status: :unauthorized,
+             layout: false
     end
   end
-
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -9,7 +9,7 @@ class ApplicationController < ActionController::Base
   before_action :set_current_user
   before_action :redirect_unauthenticated
   before_action :access_control
-  befroe_action :primary_account
+  before_action :primary_account
   layout 'application'
 
   private
@@ -69,7 +69,7 @@ class ApplicationController < ActionController::Base
   def primary_account
     if request.env['UMAPrimaryAccount'] != request.env['uid']
       render file: 'sessions/unauthorized_subsidiary',
-            status: :unauthorized
+            status: :unauthorized,
             layout: false
     end
   end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -9,6 +9,7 @@ class ApplicationController < ActionController::Base
   before_action :set_current_user
   before_action :redirect_unauthenticated
   before_action :access_control
+  befroe_action :primary_account
   layout 'application'
 
   private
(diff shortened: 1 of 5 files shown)


dfaulken:
Seems simple enough for handling the request that Shibboleth makes after people are redirected there initially. How about once people are using the site? Does this work because `nil == nil`?

AkivaGreen:
In development it works because `nil == nil`. However, in production it seems to actually work.
- vaguely quoting Matt

dfaulken:
Okay - though since some tests are what's needed to finish this PR up anyway, might be nice to write one to verify that this works when neither of those request parameters are present.

app/controllers/application_controller.rb
```diff
@@ -9,6 +9,7 @@ class ApplicationController < ActionController::Base
   before_action :set_current_user
   before_action :redirect_unauthenticated
   before_action :access_control
+  before_action :primary_account
```
dfaulken: To be fair, `:access_control` just above wasn't a great example, but it's nice to phrase methods like these as an imperative, e.g. `check_primary_account`, `handle_subsidiary_accounts`, etc.

app/views/sessions/unauthenticated_subsidiary.haml
```diff
@@ -0,0 +1,5 @@
+%h1 Login failure
+Role subsidiary accounts can not be used to access this site. Please login with your primary NetID.
```
dfaulken: `cannot` is more common usage than `can not`

dfaulken:
What do we think, @sherson? Is the text written in that warning HAML page going to be understood by people who encounter this problem? (I have absolutely no clue who these people are, besides Matt / his LastPass extension of course).

sherson:
Looks OK to me... if they don't know what a role subsidiary account is, they'll email us.

That written, I suppose we could be extra verbose:

> You are attempting to login with `uid `, which is a role subsidiary account belonging to `UMAPrimaryAccount`. Role subsidiary accounts can not be used to access this site. Please login with your primary NetID. If you need assistance, contact transit-it@admin.umass.edu.

sherson:
> I have absolutely no clue who these people are, besides Matt / his LastPass extension of course

Some examples of my subsidiary accounts: A few staff at PVTA (in case we need to grant them VPN access to reach the camera system), TPA (for their email), and several -af accounts (for student staff AD logins).

dfaulken:
Sounds good, thanks for clarifying. Couple of small changes as I suggested and this is good to go.

dfaulken:
Great, some request specs will finish this up.

dfaulken:
Also rubocop.

spec/controllers/dashboard_controller_spec.rb
```diff
@@ -119,4 +119,30 @@
       end
     end
   end
+
+  describe 'Check Primary Account' do
+    let(:admin) { create :user, :admin }
```
dfaulken: Since we only need one user per context, better to create them at that level.

dfaulken: And since there's only one test per context, you can just write

```ruby
when_current_user_is :student
```

and

```ruby
when_current_user_is :admin
```

spec/controllers/dashboard_controller_spec.rb
```diff
@@ -119,4 +119,30 @@
...
+    let(:student) { create :user, :student }
+    context 'login as subsidiary account' do
+      it 'renders unauthenticated subsidiary haml' do
+        when_current_user_is student
+        request.env['uid'] = student.id
+        request.env['UMAPrimaryAccount'] = admin.id
+        get :main
+        expect(response)
+          .to render_template 'sessions/unauthenticated_subsidiary'
+        expect(response).not_to redirect_to '/dashboard/student'
```
dfaulken: Generally nicer to use the path helper, e.g. `student_dashboard_path`.

spec/controllers/dashboard_controller_spec.rb
```diff
@@ -119,4 +119,30 @@
...
     end
   end
+
+  describe 'Check Primary Account' do
+    let(:admin) { create :user, :admin }
+    let(:student) { create :user, :student }
+    context 'login as subsidiary account' do
+      it 'renders unauthenticated subsidiary haml' do
+        when_current_user_is student
+        request.env['uid'] = student.id
```
dfaulken: `request.env.merge!` would be a little shorter here, if that's defined.

spec/controllers/dashboard_controller_spec.rb
```diff
@@ -119,4 +119,29 @@
       end
     end
   end
+
+  describe 'Check Primary Account' do
+    context 'login as subsidiary account' do
+      it 'renders unauthenticated subsidiary haml' do
+        when_current_user_is :student
+        admin = create :user, :admin
+        request.env.merge!({ 'uid' => session[:user_id],
```
dfaulken: Don't need parentheses or curly braces here.

AkivaGreen: Sadly, without the parenthesis the it won't work.

werebus: It will, but you have to drop the curly braces too.  This is because curly braces are also the way ruby syntax specifies a block argument to a method, but what you have isn't a valid block.

```ruby
request.env.merge! 'uid' => 'session[:user_id],
                   'UMAPrimaryAccount' => admin.id
```

Should be valid, I think.

AkivaGreen: It works, thanks!

PR #265, opened 2016-08-22
Unavailability calendar in pdf version of application records

### LINKED ISSUE #264: Add unavailability calendar to print view
The PDF print view doesn't include the applicant's unavailability. Time for moar Prawn!

bgregg:
^^that. Closes #264.

### DIFF
--- a/app/models/unavailability.rb
+++ b/app/models/unavailability.rb
@@ -7,6 +7,13 @@ class Unavailability < ActiveRecord::Base
   # so that they accept strings in the form of arrays
   Date::DAYNAMES.map(&:downcase).map(&:to_sym).each { |d| serialize d, Array }
 
+  # the grid method returns a data structure
+  # which represents the unavailability as a 2d array
+  # of booleans, i.e, if Sunday has unavailable times
+  # at 7AM and 8AM, then grid[0][0] and grid[0][1]
+  # will be true, and grid[0][2...14] will be false,
+  # because the applicant is available during those times.
+
   def grid
     Date::DAYNAMES.map do |name|
       daily_hours = send(name.downcase)
--- a/app/models/unavailability.rb
+++ b/app/models/unavailability.rb
@@ -6,4 +6,13 @@ class Unavailability < ActiveRecord::Base
   # this will serialize the columns of the unavailability model
   # so that they accept strings in the form of arrays
   Date::DAYNAMES.map(&:downcase).map(&:to_sym).each { |d| serialize d, Array }
+
+  def grid
+    Date::DAYNAMES.map do |name|
+      daily_hours = send(name.downcase)
+      HOURS.map do |hour|
+        daily_hours.include? hour
+      end
+    end
+  end
 end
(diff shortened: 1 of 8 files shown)


bgregg:
screen shot 2016-08-22 at 11 59 36 am

spec/factories/unavailabilities.rb
```diff
@@ -1,5 +1,12 @@
 FactoryGirl.define do
   factory :unavailability do
```
dfaulken: Does the factory need to be this complex?

app/models/unavailability.rb
```diff
@@ -6,4 +6,13 @@ class Unavailability < ActiveRecord::Base
   # this will serialize the columns of the unavailability model
   # so that they accept strings in the form of arrays
   Date::DAYNAMES.map(&:downcase).map(&:to_sym).each { |d| serialize d, Array }
+
+  def grid
```
dfaulken: I know I wrote this, but a comment explaining what type of data structure this method returns could be useful.

spec/factories/unavailabilities.rb
bgregg: Probably not. I did this as opposed to putting it directly in the spec

spec/models/application_record_spec.rb
```diff
@@ -19,6 +19,21 @@
     end
   end
 
+  describe 'unavailability_rows' do
+    before :each do
+      @record = create :application_record, :with_unavailability
+    end
+    let :call do
```
dfaulken: This could also be a one-liner.

dfaulken: As in

```ruby
let(:call){ @record.unavailability.grid }
```

spec/models/application_record_spec.rb
```diff
@@ -19,6 +19,21 @@
     end
   end
 
+  describe 'unavailability_rows' do
+    before :each do
+      @record = create :application_record, :with_unavailability
```
dfaulken: Prefer `let` blocks to instance variables.

spec/factories/unavailabilities.rb
dfaulken: Better to put it in the spec, for two reasons:

1. You don't need to read two different files to determine why a test is passing or failing;
2. A factory should only have the absolute minimum information required to create a valid instance of an object (that is, it should echo the model validations).

app/pdfs/print_record_pdf.rb
```diff
@@ -5,15 +5,15 @@ class PrintRecordPdf < Prawn::Document
   def initialize(record)
     super()
     @record = record
```
bgregg: I guess I don't really know what you mean. I see what you mean by removing that line, but then how will we access the record in methods other than initialize()?

dfaulken: In the same way that `unavailability_calendar` does, by passing it as a method argument

spec/factories/unavailabilities.rb
dfaulken: (I see you moved the logic to the spec, but it's still in this file too 👅 )

PR #268, opened 2016-09-07
between scope now actually inclusive with dates

### LINKED ISSUE #267: Applications that happened today don't show up when searching with an end date of today
An application was submitted last night at 8:49 pm, which in UTC time is this morning at 1:49 am. Don searched for applications from a week ago until today, which seems to have not included that application.

Workaround was to search with an end date of tomorrow.

Anbranin:
When giving an end_date to a record search--say it's the 7th, that would get saved as the beginning of the 7th -- midnight of the night before. Any records created on the 7th would not be included. Now it searches until the end of the day.

closes #267

### DIFF
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -14,7 +14,9 @@ class ApplicationRecord < ActiveRecord::Base
   validates :reviewed, inclusion: { in: [true, false] }
 
   scope :between,
-    -> (start_date, end_date) { where created_at: start_date..end_date.end_of_day }
+        lambda { |start_date, end_date|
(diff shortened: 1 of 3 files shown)


werebus:
Looks right to me.  Do you want to add an expectation [here](https://github.com/umts/jobapps/blob/master/spec/models/application_record_spec.rb#L72-L91)?

dfaulken:
Nice - this is exactly the kind of thing I was hoping would fix this.

Anbranin:
Added expectation. Thx @werebus

PR #270, opened 2016-09-27
Saved for later

Anbranin:
@bgregg started this, I finished it.

You can now save an application for later review:
screen shot 2016-09-27 at 3 44 40 pm
If you click the box that says "email to applicant" the note will get emailed to the applicant. 
Selecting a date does this: A cron job runs every 1.day to move any records with an expired date_for_later back onto the dashboard.
Link to view the saved applications here:
screen shot 2016-09-27 at 3 45 22 pm
Which goes to here:
screen shot 2016-09-27 at 3 45 57 pm
There's also a button to move the record back to the dashboard if you wish.


app/views/jobapps_mailer/application_denial.text.erb
```diff
@@ -1,7 +1,7 @@
 Hello <%= @user.first_name %>,
 
 <% if configured_value [:on_application_denial, :notify_of_reason], default: true %>
-  Your application to the <%= @application_record.position.try(:name) %> positon has been denied due to the following reason:
+  Your application to the <%= @application_record.position.try(:name) %> position has been denied due to the following reason:
   <%= @application_record.staff_note %>
```
Anbranin: Just a spelling error here

Anbranin:
can someone review this? 😢

app/controllers/dashboard_controller.rb
```diff
@@ -13,7 +13,8 @@ def main
   def staff
     @departments = Department.includes :positions
     @pending_interviews = Interview.pending.group_by(&:position)
-    @pending_records = ApplicationRecord.pending.newest_first
+    @pending_records = ApplicationRecord.where(saved_for_later: false)
```
akaplo: Should we also provide a way to see the applications that have been saved for later?  I see something to that effect in the `PositionsController`, but perhaps a dashboard option or link might work too?

Anbranin: There is a link on the dashboard, see screenshot above.

app/mailers/jobapps_mailer.rb
```diff
@@ -42,6 +42,16 @@ def interview_reschedule(interview)
          reply_to: reply_to
   end
 
+  def send_note_for_later(application_record)
```
akaplo: What about the `note_for_later`?  Aren't we sending that along with the email if it exists?

app/models/application_record.rb
```diff
@@ -75,6 +75,23 @@ def pending?
...
 
+  def save_for_later(date = nil, note = nil, mail = false)
+    update_attributes(saved_for_later: true,
+                      date_for_later: date,
+                      note_for_later: note)
+    JobappsMailer.send_note_for_later(self).deliver_now if mail
+  end
+
+  def unsave
+    update_attributes(saved_for_later: false,
```
akaplo: Perhaps we should also clear out `note_for_later`.  What happens if we unsave and then resave?

app/controllers/application_records_controller.rb
```diff
@@ -66,6 +66,22 @@ def review
     show_message :application_review,
                  default: 'Application has been marked as reviewed.'
     @record.update reviewed: true
+    @record.update saved_for_later: false
```
Anbranin: Absolutely they can.

app/mailers/jobapps_mailer.rb
Anbranin: It's an attribute of record. The line /@record = application_record allows me to use any attributes or methods of that application record in the mailer view, which you will see when you get to that portion.

app/models/application_record.rb
Anbranin: No. The note can be replaced if necessary. Also, say you had saved an application in May for later review in September, and it got moved back to the dashboard. Would it not be helpful to see your note about why you'd saved it?

app/views/application_records/show.haml
```diff
@@ -6,6 +6,10 @@ by:
 = @record.user.name_and_email
 = button_to 'Print this page',
   application_record_path(@record, format: :pdf), method: :get
+- if !@record.saved_for_later && @record.note_for_later.present?
```
akaplo: If an app has recently been unsaved but didn't have a note, perhaps we could add something similar to this, but that doesn't say `Your note:`

app/mailers/jobapps_mailer.rb
akaplo: How about a mailer method that emails the Staff user tied to this Position when a application comes out of `saved_for_later`?  This might be a question for Don as well.

app/views/application_records/show.haml
Anbranin: Meh

Anbranin: You're right

Anbranin: But there's no way of saying "if it was previously saved for later." You can't query an object based on an attribute they previously had.

db/schema.rb
```diff
@@ -22,15 +22,18 @@
...
     t.boolean  "reviewed"
-    t.integer  "position_id", limit: 4
-    t.text     "staff_note",  limit: 65535
-    t.string   "ethnicity",   limit: 255
-    t.string   "gender",      limit: 255
+    t.integer  "position_id",     limit: 4
+    t.text     "staff_note",      limit: 65535
+    t.string   "ethnicity",       limit: 255
+    t.string   "gender",          limit: 255
+    t.boolean  "saved_for_later",               default: false
```
akaplo: Awkward spacing on purpose?

Anbranin: I don't write the schema, it writes itself

app/mailers/jobapps_mailer.rb
akaplo: Yep, gotcha.

app/models/application_record.rb
akaplo: #truth

app/views/application_records/show.haml
akaplo: That's a good point.  Is it worth adding the attribute for this one thing/do you think it could be useful in extending this feature someday?

app/mailers/jobapps_mailer.rb
Anbranin: Not for now, want MVP. Maybe later. Good thought. Don himself has expressed no interest.

dfaulken: @akaplo I agree that's a good thought - feel free to open an issue about it referencing this PR.

app/views/application_records/show.haml
dfaulken: > You can't query an object based on an attribute they previously had.

Not entirely true, we could enable `paper_trail`.

But of course, isn't the presence of a `note_for_later` sufficient to assume that it was once saved for later?

Gemfile
```diff
@@ -18,6 +18,7 @@ gem 'redcarpet'
 gem 'sass-rails'
 gem 'snappconfig'
 gem 'uglifier'
+gem 'whenever', require: false
```
dfaulken: Interesting, why `require: false`?

app/controllers/application_records_controller.rb
```diff
@@ -66,6 +66,22 @@ def review
...
+    redirect_to staff_dashboard_path
+  end
+
+  def toggle_saved_for_later
+    if @record.saved_for_later
+      @record.unsave
+      flash[:message] = 'Application moved back to dashboard.'
+    else
+      date = if params[:date_for_later].present?
```
dfaulken: So otherwise, `date` will be `nil`, right?

Will that cause problems?

app/controllers/application_records_controller.rb
```diff
@@ -66,6 +66,22 @@ def review
...
+               Date.strptime(params[:date_for_later], '%m/%d/%Y')
+             end
+      mail = true if params[:mail_to_applicant] == '1'
```
dfaulken: Interesting - what's your reasoning for this being an individually configurable choice rather than a globally configurable one?

app/controllers/positions_controller.rb
```diff
@@ -16,6 +19,10 @@ def destroy
     redirect_to staff_dashboard_path
   end
 
+  def saved_applications
+    @saved = ApplicationRecord.where(saved_for_later: true, position: @position)
```
dfaulken: Given that we use it a couple of times, perhaps a `saved_for_later` scope?

app/mailers/jobapps_mailer.rb
```diff
@@ -42,6 +42,16 @@ def interview_reschedule(interview)
...
+    template = application_record.position.application_template
+    @user = application_record.user
+    @record = application_record
+    reply_to = template.try :email
+    mail to: application_record.user.email,
```
dfaulken: Can write just `@user.email` here

app/models/application_record.rb
```diff
@@ -77,6 +77,23 @@ def pending?
     !reviewed
   end
 
+  def save_for_later(date = nil, note = nil, mail = false)
```
dfaulken: Seems like keyword arguments might enhance readability here.

app/models/application_record.rb
```diff
@@ -77,6 +77,23 @@ def pending?
...
+    update_attributes(saved_for_later: true,
+                      date_for_later: date,
+                      note_for_later: note)
+    JobappsMailer.send_note_for_later(self).deliver_now if mail
+  end
+
+  def unsave
```
dfaulken: I think given that this does a very specific thing, rather than just generally undoing changes made on an object, I'd prefer a more specific name. Perhaps `move_to_front` or `move_to_dashboard`?

dfaulken: That is, the method name sort of implies that it's more generic than it actually is.

app/views/application_records/show.haml
dfaulken: In general, prefer calling boolean attributes as predicate methods - here, `!@record.saved_for_later?`

dfaulken: And the same in a few other places in this PR.

app/views/application_records/show.haml
```diff
@@ -65,6 +69,40 @@ by:
                          required: true,
                          value: @record.position.default_interview_location
       .note_item= submit_tag 'Review application and schedule interview'
+    - if !@record.saved_for_later
```
dfaulken: You have "if not X" followed by "if X". Seems like "if X" and "else", with the respective code reversed, would be simpler.

app/views/jobapps_mailer/send_note_for_later.text.erb
```diff
@@ -0,0 +1,9 @@
+Hello <%= @user.first_name %>,
+
+Your application to the <%= @record.position.try(:name) %> position has been saved for later review due to the following reason:
+
+<%= @record.note_for_later %> 
+
+Best,
+
+<%= configured_value [:organization_name] %>
```
dfaulken: Perhaps something to indicate that this is an automated email?

app/views/positions/saved_applications.haml
```diff
@@ -0,0 +1,18 @@
+%table.data_table.display#main_data_table
+  %thead
+    %tr
+      %th Date Submitted
+      %th Name
+      %th Note
+      %th Date To Review
```
dfaulken: Stylistically I usually prefer not capitalizing these; here, `Date to review`.

Either way, though, it's incorrect to capitalize prepositions.

app/views/positions/saved_applications.haml
```diff
@@ -0,0 +1,18 @@
...
+    - @saved.each do |record|
+      %tr
+        %td{ data: { order: record.created_at.to_i } }
+          = format_date_time record.created_at
+        %td
+          = link_to(record.user.proper_name, application_record_path(record.id))
+        %td
+          = record.note_for_later
+        %td
+          = format_date record.date_for_later if record.date_for_later
```
dfaulken: If it's `present?` is what we're concerned with, more natural to write it like that

config/schedule.rb
```diff
@@ -0,0 +1,4 @@
+every :day do
```
dfaulken: At what time will this occur?

lib/date_and_time_methods.rb
```diff
@@ -7,6 +7,10 @@ def format_date_time(datetime, options = {})
     end.squish
   end
 
+  def format_date(date)
```
dfaulken: Tough to know from the PHP format escape chars what this will look like - perhaps an example in a comment might be nice.

I ask because it seems like there are too many commas here, but I'm not sure

spec/controllers/application_records_controller_spec.rb
```diff
@@ -322,6 +323,28 @@
...
 
+  describe 'POST #toggle_saved_for_later' do
+    before :each do
+      when_current_user_is :staff
+    end
+    context 'record not saved for later' do
+      let!(:record) { create :application_record, saved_for_later: false }
+      let(:submit) { post :toggle_saved_for_later, id: record.id }
+      it 'calls record.save_for_later' do
+        expect_any_instance_of(ApplicationRecord).to receive(:save_for_later)
```
dfaulken: Does it not work here to establish a mock on the specific instance you create on line 331?

spec/features/application_record_save_for_later_spec.rb
```diff
@@ -0,0 +1,80 @@
+require 'rails_helper'
```
dfaulken: In general, we should be consistent with `eq` and `eql`. I've been using `eql` - I'm open to arguments for using one or the other, but I think not both would be good.

spec/features/application_record_save_for_later_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+    end
+    it 'puts a notice in the flash' do
+      click_button 'Move back to dashboard'
+      expect(page).to have_text 'Application moved back to dashboard'
+    end
+  end
+  context 'adding a note and date for later' do
+    let!(:record) { create :application_record, reviewed: false }
+    before :each do
+      visit application_record_url(record)
```
dfaulken: Why is this setup preferable to just creating the application record as it should be and then visiting the page?

spec/mailers/jobapps_mailer_spec.rb
```diff
@@ -145,6 +145,39 @@
     end
   end
 
+  describe 'send_note_for_later' do
+    let(:record) do
+      create :application_record,
+             saved_for_later: true,
+             note_for_later: 'We need you to grow up a little'
+    end
+    let!(:template) do
```
dfaulken: I don't see that you actually use this variable - perhaps I'm mistaken

spec/models/application_record_spec.rb
```diff
@@ -184,6 +184,81 @@
...
+    context 'mail to applicant not desired' do
+      it 'does not call the mailer method' do
+        expect(JobappsMailer).not_to receive(:send_note_for_later)
+        record.save_for_later(nil, nil, false)
+      end
+    end
+  end
+
+  describe 'unsave' do
+    let(:record) do
```
dfaulken: You alternate a fair bit in this PR between

```ruby
let :record do
```

and 

```ruby
let(:record) do
```

It'd be good to be consistent.

app/views/application_records/show.haml
akaplo: Yes, but there's a case where it was saved for later, but there was no note given.  In this case, `note_for_later` will always have been `nil`.

app/models/application_record.rb
akaplo: Agreed on the name.  I almost commented here, but couldn't think of anything more constructive than "I don't like it"

app/views/application_records/show.haml
dfaulken: So I'm confused - what's the problem you're asking about @akaplo?

akaplo: > isn't the presence of a note_for_later sufficient to assume that it was once saved for later?

We're talking about two different things now, I think.  

When an application is saved for later, it need not have a `note`.  This block of code marks an activate application that was at one point `saved_for_later`, but is reliant on the fact that it included a `note`.  I wanted to be able to say "the application was saved for later" regardless of whether it ever had a `note`.  If it didn't have a `note` and currently is _not_ `saved_for_later`, there's no way (except perhaps with `paper_trial`, like you said), to know whether it was saved for later.

So, the presence of a `note_for_later` *is* enough to assume it was once `saved_for_later`, but this would cause us to miss apps that were saved but didn't have a `note.`  Is that clearer?

dfaulken: Right, I agree.

So

```
This app was saved for future review.
- if theres_a_note
  = the note
```

Gemfile
Anbranin: That's what the docs told me to do. My understanding is limited, but you require: false whenever you have something that runs from the command line or as a cron job but is not a library used in the code itself.

app/controllers/application_records_controller.rb
Anbranin: No, no problems if date for later is nil. It is not called when it could be nil.

app/controllers/application_records_controller.rb
Anbranin: Because you might want applicants to see some notes, and maybe you don't want them to see others. I think actually that this is preferable. Perhaps the default should be a checked box, but that'll be after Don gives it a test run.

Gemfile
dfaulken: Makes sense.

app/controllers/application_records_controller.rb
dfaulken: Great!

app/mailers/jobapps_mailer.rb
Anbranin: Oh wow that's like 3 lines above it lol

app/controllers/application_records_controller.rb
dfaulken: OK - do you have an actual use case in mind, or is this just your intuition?

app/controllers/positions_controller.rb
Anbranin: We use it once, so no

Anbranin: Good thought tho

app/models/application_record.rb
Anbranin: Oh ya

Anbranin: K forgot about those forgot they r awesome wheeee

app/models/application_record.rb
Anbranin: I honestly do not know why I didn't choose 'move_to_dashboard' given that there's a cron job of almost the same name.

app/views/application_records/show.haml
Anbranin: My brain and logic trees sometimes makes me 😵

app/views/positions/saved_applications.haml
Anbranin: Quite right. I like capitalization of titles. You do not. But neither of us like capitalizations of prepositions.

app/models/application_record.rb
dfaulken: :-D

app/views/jobapps_mailer/send_note_for_later.text.erb
Anbranin: You know what, I'm gonna add that in later with the email-rehaul.

config/schedule.rb
Anbranin: NO IDEA

Anbranin: 12AM now. I specified.

lib/date_and_time_methods.rb
Anbranin: screen shot 2016-10-03 at 3 36 09 pm

spec/controllers/application_records_controller_spec.rb
Anbranin: It does not.

spec/features/application_record_save_for_later_spec.rb
Anbranin: fine. They are slightly different, but lemme change em, I got no opinion

spec/controllers/application_records_controller_spec.rb
dfaulken: Weird - haven't we found that that has worked under other circumstances?

lib/date_and_time_methods.rb
dfaulken: Right, so yes, there are

app/views/application_records/show.haml
Anbranin: Someday? Sure, but I don't write apps for someday. I make them for now, but future-proofed. 
It would be easy enough to add that attribute, but as it stands currently, there's always a note. If there's no note, someone would just have an apparently new application to read through. Not a big deal--they'll either come to the same conclusion they came to before, or they'll reject it based on new information.

lib/date_and_time_methods.rb
Anbranin: Now there are not thx

spec/controllers/application_records_controller_spec.rb
Anbranin: It's more like I could have sworn it worked but couldn't remember any specific instances where it did. It just doesn't do the thing.

dfaulken: Ah, okay

app/controllers/application_records_controller.rb
Anbranin: Intuition--but I told Don how the app currently stands and he likes it the check-the-box idea. I could definitely _imagine_ some scenarios, but so could you I'm sure.

app/views/application_records/show.haml
dfaulken: @Anbranin and I talked and decided that requiring a note is the simplest way to resolve this.

Anbranin: WELL we have decided to REQUIRE the note. :)

Anbranin: Because, @akaplo, you make an excellent point.

Anbranin: YUP THIS IS IT THIS IS HOW YOU DO IT
screen shot 2016-10-03 at 4 34 09 pm

app/controllers/application_records_controller.rb
dfaulken: Could you share an actual use case with the class?

spec/features/application_record_save_for_later_spec.rb
Anbranin: Because! I wanted to actually "fill_in" the fields, making sure they are there. Otherwise it's more like a unit test.

spec/mailers/jobapps_mailer_spec.rb
Anbranin: I needed to create a template so I had an email. The factory doesn't have one by default because it is not a required attribute.

spec/models/application_record_spec.rb
Anbranin: Any instances when I left the parentheses in was by mistake.

spec/mailers/jobapps_mailer_spec.rb
dfaulken: Sounds good - then you can either integrate this into your `before :each` or reference it directly as `template` in the test rather than indirectly through other objects.

spec/features/application_record_save_for_later_spec.rb
dfaulken: We discussed in person how this functionality should be put into a separate test, that is:

1. Test that you can get an object to point X.
2. Test that, given an object at point X, assertions Y, Z... are true.

app/views/positions/saved_applications.haml
dfaulken: Neither of us *likes*, I think you mean

app/views/jobapps_mailer/send_note_for_later.text.erb
dfaulken: Sounds good - let's open up an issue for that if we haven't already

Anbranin:
@dfaulken I think I fixed all the things with latest push.

dfaulken:
Nice work 😄

PR #272, opened 2016-10-12
Ensure User Creation Has '@umass.edu' in SPIRE Field

### LINKED ISSUE #266: Desired SPIRE value unclear when creating user
In the users/new form, (used by admins to create a new staff user), the SPIRE field should be /spire_number@umass.edu, not merely the number. However, the admin creating a user yesterday did not know this and merely entered the number, and the new staff user could not log in until I changed his spire ID in the database to have an /@umass.edu after. Could we somehow make it clear that the number should be entered with a /@umass.edu after it? Perhaps a placeholder?

akaplo:
Closes #266.  Adds a placeholder to the SPIRE field on `Users#new` and adds a validation in the User model to check that this field receives 8 digits followed by '@umass.edu'.  Previously, we only validated `presence` and `uniqueness`.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -11,7 +11,7 @@ class User < ActiveRecord::Base
             presence: true
   validates :staff, inclusion: { in: [true, false],
                                  message: 'must be true or false' }
-  validates :spire, uniqueness: true
+  validates :spire, uniqueness: true, format: {with: /@umass\.edu\z/}
 
   default_scope { order :last_name, :first_name }
   scope :staff,    -> { where staff: true }
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -11,7 +11,7 @@ class User < ActiveRecord::Base
             presence: true
   validates :staff, inclusion: { in: [true, false],
                                  message: 'must be true or false' }
-  validates :spire, uniqueness: true, format: {with: /@umass\.edu\z/}
+  validates :spire, uniqueness: true, format: {with: /\d{8}@umass\.edu/}
 
   default_scope { order :last_name, :first_name }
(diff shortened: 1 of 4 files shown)


app/views/users/_form.haml
```diff
@@ -11,7 +11,7 @@
     = f.text_field :email, required: true
   .form-group.row
     .col-md-2= f.label :spire
-    = f.text_field :spire, required: true
+    = f.text_field :spire, required: true, placeholder: 'Spire-Id@umass.edu'
```
dfaulken: I usually put examples in placeholders.

`placeholder: 'e.g. 08675309@umass.edu'`

app/models/user.rb
```diff
@@ -11,7 +11,7 @@ class User < ActiveRecord::Base
             presence: true
   validates :staff, inclusion: { in: [true, false],
                                  message: 'must be true or false' }
-  validates :spire, uniqueness: true
+  validates :spire, uniqueness: true, format: { with: /\A\d{8}@umass\.edu\z/ }
```
dfaulken: Oh good point — would you mind opening an issue (or PR) making sure we're checking beginning and end of string on screaming-dinosaur?

app/views/users/_form.haml
akaplo: Fair point.

app/models/user.rb
dfaulken: Thanks!

dfaulken:
Do we have any integration tests concerning creating users?

akaplo:
[We do!](https://github.com/umts/jobapps/blob/master/spec/features/user_creation_spec.rb#L44)

PR #274, opened 2016-11-02
Saved application

### LINKED ISSUE #271: Alert Staff That a Saved Application Has Come Back
First thought of in #270.  When we save an Application for a Position for later, it may be useful to email the staff user tied to that Position when the application has been auto-removed from the "saved for later" list and brought back into the Dashboard.  This could be useful for staff users who don't check their dashboard every day so that they can review an application in a timely manner.

AkivaGreen:
I would like to close #271 with this pull request.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,5 +27,3 @@ coverage/assets/*
 coverage/.resultset.*
 
 Guardfile
-
-.byebug_history
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,5 @@ coverage/assets/*
 coverage/.resultset.*
 
 Guardfile
+
+.byebug_history
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -19,3 +19,11 @@
 //= require multiple-select
 //= require jquery.maskedinput
 //= require_tree .
+$(function(){
+  $('form').on('change', '#date_for_later', function(){
+    if(this.value.length > 0){
+      $('#notification').removeClass('hidden');
+    }
+    else $('#notification').addClass('hidden');
+  });
+});
--- a/app/controllers/application_records_controller.rb
+++ b/app/controllers/application_records_controller.rb
@@ -133,10 +133,10 @@ class ApplicationRecordsController < ApplicationController
 
   def modify_params
     parameters = {}
-    parameters[:date] = if params[:date_for_later].present?
-                          Date.strptime(params[:date_for_later], '%m/%d/%Y')
-                        end
-    parameters[:mail] = true if params[:mail_to_applicant] == '1'
+    if params[:date_for_later].present?
+      parameters[:date] = Date.strptime(params[:date_for_later], '%m/%d/%Y')
+    end
+    parameters[:mail] = params[:mail_to_applicant] == '1'
     parameters
   end
 end
--- a/app/controllers/application_records_controller.rb
+++ b/app/controllers/application_records_controller.rb
@@ -81,7 +81,8 @@ class ApplicationRecordsController < ApplicationController
       mail = true if params[:mail_to_applicant] == '1'
       @record.save_for_later(date: date,
                              note: params[:note_for_later],
-                             mail: mail)
+                             mail: mail,
+                             email: params[:email_to_notify])
       flash[:message] = 'Application saved for later.'
     end
     redirect_to staff_dashboard_path
--- a/app/controllers/application_records_controller.rb
+++ b/app/controllers/application_records_controller.rb
@@ -75,13 +75,10 @@ class ApplicationRecordsController < ApplicationController
       @record.move_to_dashboard
       flash[:message] = 'Application moved back to dashboard.'
     else
-      date = if params[:date_for_later].present?
-               Date.strptime(params[:date_for_later], '%m/%d/%Y')
-             end
-      mail = true if params[:mail_to_applicant] == '1'
-      @record.save_for_later(date: date,
+      parameters = modify_params
+      @record.save_for_later(date: parameters[:date],
                              note: params[:note_for_later],
-                             mail: mail,
+                             mail: parameters[:mail],
                              email: params[:email_to_notify])
       flash[:message] = 'Application saved for later.'
     end
@@ -132,5 +129,14 @@ class ApplicationRecordsController < ApplicationController
 
   def record_params
     params.permit(:position_id, :ethnicity, :gender)
+  end
+
+  def modify_params
+    parameters = {}
+    parameters[:date] = if params[:date_for_later].present?
+                          Date.strptime(params[:date_for_later], '%m/%d/%Y')
+                        end
+    parameters[:mail] = true if params[:mail_to_applicant] == '1'
+    parameters
   end
 end
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -95,7 +95,8 @@ class ApplicationRecord < ActiveRecord::Base
     records.each do |record|
       record.move_to_dashboard
       record.position.subscriptions.each do |sub|
-        JobappsMailer.saved_application_notification(sub, sub.position, record.user)
+        JobappsMailer.saved_application_notification(sub, sub.position,
+                                                     record.user)
       end
     end
   end
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -92,29 +92,14 @@ class ApplicationRecord < ActiveRecord::Base
 
   def self.move_to_dashboard
     records = where('date_for_later <= ?', Time.zone.today)
-    emails = {}
-    query = { notification: true }
     if records.count == 1
       record = records.first
-      record.position.subscriptions.where(query).each do |sub|
+      record.position.subscriptions.notification.each do |sub|
         JobappsMailer.saved_application_notification(sub, sub.position, record)
       end
-    else 
-      records.each do |record|
-        record.position.subscriptions.where(query).each do |sub|
-          if emails[sub.email][sub.position.name].present?
-            emails[sub.email][sub.position.name].push(record)
-          else
-            emails[sub.email][sub.position.name] = [record]
-          end
-        end
-        record.move_to_dashboard
-      end
-      if emails.present?
-        emails.each_key do |email|
-          JobappsMailer.saved_applications_notification(emails[email], email)
-        end
-      end
+      record.move_to_dashboard
+    else
+      notification_emails(records)
     end
   end
 
@@ -158,6 +143,25 @@ class ApplicationRecord < ActiveRecord::Base
     all_ethnicities.map do |ethnicity|
       specific_records = ethnicity_records.where ethnicity: ethnicity
       [ethnicity, specific_records.count, specific_records.interview_count]
+    end
+  end
+
+  def notification_emails(records)
+    emails = {}
+    records.each do |record|
+      record.position.subscriptions.notification.each do |sub|
+        if emails[sub.email][sub.position.name].present?
+          emails[sub.email][sub.position.name].push(record)
+        else
+          emails[sub.email][sub.position.name] = [record]
+        end
+      end
+      record.move_to_dashboard
+    end
+    if emails.present?
+      emails.each_key do |email|
+        JobappsMailer.saved_applications_notification(emails[email], email)
+      end
     end
   end
 end
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -93,16 +93,28 @@ class ApplicationRecord < ActiveRecord::Base
   def self.move_to_dashboard
     records = where('date_for_later <= ?', Time.zone.today)
     emails = {}
-    records.each do |record|
-      query = { notification: true }
+    query = { notification: true }
+    if records.count == 1
+      record = records.first
       record.position.subscriptions.where(query).each do |sub|
-        if emails["#{sub.email}"]["#{sub.position}"].present?
-          emails["#{sub.email}"]["#{sub.position}"].push(record.user)
-        else
-          emails["#{sub.email}"]["#{sub.position}"] = [record.user]
+        JobappsMailer.saved_application_notification(sub, sub.position, record)
+      end
+    else 
+      records.each do |record|
+        record.position.subscriptions.where(query).each do |sub|
+          if emails[sub.email][sub.position.name].present?
+            emails[sub.email][sub.position.name].push(record)
+          else
+            emails[sub.email][sub.position.name] = [record]
+          end
+        end
+        record.move_to_dashboard
+      end
+      if emails.present?
+        emails.each_key do |email|
+          JobappsMailer.saved_applications_notification(emails[email], email)
         end
       end
-      record.move_to_dashboard
     end
   end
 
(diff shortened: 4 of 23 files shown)


AkivaGreen:
I still need to adjust the test to bring coverage back up

app/models/application_record.rb
```diff
@@ -92,7 +92,13 @@ def move_to_dashboard
 
   def self.move_to_dashboard
     records = where('date_for_later <= ?', Time.zone.today)
-    records.each(&:move_to_dashboard)
+    records.each do |record|
+      record.move_to_dashboard
+      record.position.subscriptions.each do |sub|
```
dfaulken: Hmm, seems like this could be handled in the instance `move_to_dashboard` method (and then this could be put back the way it was).

dfaulken: No — I take that back! It would be *best* not to spam someone with 7 emails if 7 applications are reactivated for the same position.

Here's what we should do:

1. For all applications which will be reactivated, find the email addresses of the subscriptions.
2. Create a hash mapping each email address to the reactivated applications about which the person needs to be notified.
3. Send a single email which breaks down how many applications were reactivated per position.

Note that a single reactivated application could be sent to multiple people, one of whom might also have a subscription to another position — so there is a non-trivial amount of data processing to do here.

app/views/jobapps_mailer/saved_application_notification.text.erb
```diff
@@ -0,0 +1,5 @@
+Hello,
+
+A saved for later application has been moved back to dashboard.
```
dfaulken: This sentence is a little awkward. Perhaps

```
An application which you had saved for later has been reactivated and is now visible on the dashboard (link to dashboard).
```

.gitignore
```diff
@@ -27,3 +27,5 @@ coverage/assets/*
 coverage/.resultset.*
 
 Guardfile
+
+.byebug_history
```
dfaulken: What's this?

db/migrate/20161103211510_add_notification_to_application_record.rb
```diff
@@ -0,0 +1,5 @@
+class AddNotificationToApplicationRecord < ActiveRecord::Migration
```
dfaulken: Now that you've run them yourself, you can remove the two migrations adding and removing the column to/from the wrong table.

app/models/application_record.rb
```diff
@@ -92,12 +92,17 @@ def move_to_dashboard
 
   def self.move_to_dashboard
     records = where('date_for_later <= ?', Time.zone.today)
+    emails = {}
     records.each do |record|
-      record.move_to_dashboard
-      record.position.subscriptions.each do |sub|
-        JobappsMailer.saved_application_notification(sub, sub.position,
-                                                     record.user)
+      query = { notification: true }
```
dfaulken: I think an ActiveRecord scope would be a nicer approach.

app/models/application_record.rb
```diff
@@ -92,12 +92,17 @@ def move_to_dashboard
...
+      record.position.subscriptions.where(query).each do |sub|
+        if emails["#{sub.email}"]["#{sub.position}"].present?
```
dfaulken: You don't need to string interpolate a string.

.gitignore
dfaulken: Yes, please delete this.

dfaulken:
@AkivaGreen what's the status of this? Looks like you've done some work since, but Travis is still failing.

AkivaGreen:
@dfaulken, It is in need of some more work and then tests for the new functionality. I should have this completely done by the time thanksgiving break is over.

dfaulken:
Sounds great, thanks for the update.

AkivaGreen:
Travis and CodeClimate are passing all that is left is some Javascript and Tests to bring coverage back up to 100%

app/models/application_record.rb
```diff
@@ -98,7 +98,7 @@ def self.move_to_dashboard
         JobappsMailer.saved_application_notification(sub, sub.position, record)
       end
       record.move_to_dashboard
-    else
+    elsif records.count > 1
```
dfaulken: http://api.rubyonrails.org/classes/Enumerable.html#method-i-many-3F

AkivaGreen:
This is ready for another review

app/assets/javascripts/application.js
```diff
@@ -19,3 +19,11 @@
 //= require multiple-select
 //= require jquery.maskedinput
 //= require_tree .
+$(function(){
+  $('form').on('change', '#date_for_later', function(){
+    if(this.value.length > 0){
```
dfaulken: A few things here:

1. It's best to avoid putting javascript in the application.js file. Since this pertains to a specific view, let's create another file for it and put this there. Keep in mind that this will be CoffeeScript.
2. Prefer `$(this).val()` to `this.value`. In general, prefer `$(this)` in jQuery, for reasons which are complex but I can go into if you like.
3. For removing or adding a class based on a boolean value, there's always `toggleClass`, which with 1 argument will toggle it based on its present value, but with 2 will use the boolean argument to determine whether the class should be present or not.

All that to say, we end up with:

```coffee
$('form').on 'change', '#date_for_later', ->
  $('#notification').toggleClass 'hidden', $(this).val().length > 0
  return
```

app/controllers/application_records_controller.rb
```diff
@@ -132,4 +130,13 @@ def given_or_all_department_ids
   def record_params
     params.permit(:position_id, :ethnicity, :gender)
   end
+
+  def modify_params
```
dfaulken: This could use a more specific name. `modify_params` doesn't offer too much information about what it does or to what controller method(s) it's applicable. I can offer some suggestions if you need.

app/controllers/application_records_controller.rb
```diff
@@ -132,4 +130,13 @@ def given_or_all_department_ids
...
+    parameters = {}
+    parameters[:date] = if params[:date_for_later].present?
+                          Date.strptime(params[:date_for_later], '%m/%d/%Y')
+                        end
+    parameters[:mail] = true if params[:mail_to_applicant] == '1'
```
dfaulken: This line can be shortened to `parameters[:mail] = params[:mail_to_applicant] == 1`

app/controllers/application_records_controller.rb
```diff
@@ -132,4 +130,13 @@ def given_or_all_department_ids
   def record_params
     params.permit(:position_id, :ethnicity, :gender)
   end
+
+  def modify_params
+    parameters = {}
+    parameters[:date] = if params[:date_for_later].present?
```
dfaulken: Since it won't be set otherwise, for the sake of readability I might suggest leading with the conditional, then doing the assignment if the condition is true.

app/models/application_record.rb
```diff
@@ -78,10 +78,11 @@ def pending?
     !reviewed
   end
 
-  def save_for_later(date: nil, note: nil, mail: false)
+  def save_for_later(date: nil, note: nil, mail: false, email: nil)
```
dfaulken: Nice use of kwargs here.

app/models/application_record.rb
```diff
@@ -92,6 +93,15 @@ def move_to_dashboard
 
   def self.move_to_dashboard
     records = where('date_for_later <= ?', Time.zone.today)
+    email_records = where('date_for_later <= ?', Time.zone.today)
+                    .where("email_to_notify is NOT NULL and
```
dfaulken: Could use Ruby syntax here:

```ruby
.where.not(email_to_notify: [nil, ''])
```

app/models/application_record.rb
```diff
@@ -92,6 +93,15 @@ def move_to_dashboard
...
+                            email_to_notify != ''")
+    if email_records.count == 1
```
dfaulken: Just as there's a `many?` there's also a `one?`

app/models/application_record.rb
```diff
@@ -137,4 +147,29 @@ def self.ethnicity_eeo_data(records)
       [ethnicity, specific_records.count, specific_records.interview_count]
     end
   end
+
+  def self.notification_emails(records)
+    emails = Hash.new { |h, k| h[k] = {} }
```
dfaulken: If you're going to call a method including the word `build` on it, seems weird to have to initialize first.

app/models/application_record.rb
```diff
@@ -137,4 +147,29 @@ def self.ethnicity_eeo_data(records)
...
+    records.each do |record|
+      emails = build_email_hash(emails, record)
+    end
+    send_emails(emails)
+  end
+
+  def self.send_emails(emails)
```
dfaulken: The word `email` is way too prevalent here. You could help fix this by iterating with `each_pair` and then assigning a meaningful name to the value, since you do call `emails[email]` in the loop.

app/models/application_record.rb
```diff
@@ -137,4 +147,29 @@ def self.ethnicity_eeo_data(records)
...
+    if emails.present?
+      emails.each_key do |email|
+        JobappsMailer.saved_applications_notification(emails[email], email)
+      end
+    end
+  end
+
+  def self.build_email_hash(emails, record)
```
dfaulken: I think this method could use a comment explaining what it returns given its inputs.

app/models/subscription.rb
```diff
@@ -7,4 +7,6 @@ class Subscription < ActiveRecord::Base
   validates :email,
             :user,
             :position, presence: true
+
+  scope :notification, -> { where(notification: true) }
```
dfaulken: Scope names should generally be an adjective or some kind of modifying prepositional phrase.

app/views/application_records/show.haml
```diff
@@ -100,6 +100,10 @@ by:
           %em
             If a date is selected, the application will move back onto the
             dashboard at that time.
+        %p#notification.hidden
+          %em
+            Get notified when this happens: 
+          = text_field_tag :email_to_notify, '', placeholder: 'email'
```
dfaulken: Placeholders should generally be an example of what will go in the field, so something like `joe@schmo.com` would be good here.

app/views/jobapps_mailer/saved_applications_notification.text.erb
```diff
@@ -0,0 +1,10 @@
+Hello,
+
+The following applications have been reactivated:
+<% @info.each_key do |position| %>
```
dfaulken: Here too, if you need the value (which you do), better to iterate with `each_pair`.

It's surprising to me that Rubocop doesn't have a cop for this. [Perhaps we can write one of our own](https://github.com/umts/custom-cops).

app/views/jobapps_mailer/saved_applications_notification.text.erb
```diff
@@ -0,0 +1,10 @@
...
+  Position: <%= position %>
+  <% @info[position].each do |record| %>
+    Applicant: <%= record.user.full_name %>,
+    Date Applied: <%= record.created_at.strftime('%m/%d/%Y %H:%M') %>
```
dfaulken: Given that the value is a date, I think just `Applied:` or `Applied on:` will be perfectly clear, though if you disagree I don't have strong feelings about this.

spec/controllers/application_records_controller_spec.rb
```diff
@@ -338,8 +338,8 @@
     context 'record previously saved for later' do
       let!(:record) do
         create :application_record,
-          saved_for_later: true,
-          note_for_later: 'this needs to be here'
+               saved_for_later: true,
```
dfaulken: Hmm, did Rubocop do this or did you?

spec/mailers/jobapps_mailer_spec.rb
```diff
@@ -210,4 +210,61 @@
...
+      JobappsMailer.saved_application_notification @record
+    end
+    it 'emails to the email_to_notify value' do
+      expect(output.to).to eql Array('foo@example.com')
+    end
+    it 'has a subject that includes the words Saved application' do
+      expect(output.subject).to include 'Saved application'
+    end
+    it 'includes the position name' do
+      expect(output.body.encoded).to include @record.position.name
```
dfaulken: Interesting, what does `encoded` do to `body.output`?

spec/mailers/jobapps_mailer_spec.rb
```diff
@@ -210,4 +210,61 @@
...
+      JobappsMailer.saved_applications_notification @info, @email
+    end
+    it 'emails from default configured value' do
+      expect(output.to).to eql Array(@email)
+    end
+    it 'has a subject that includes the words Saved applications' do
+      expect(output.subject).to include 'Saved applications'
+    end
+    it 'includes the position names' do
+      expect(output.body.encoded).to include @info['Bus'].first.position.name
```
dfaulken: Seems simpler to define the application records as instance variables and use them in your expectations here and below.

spec/models/application_record_spec.rb
```diff
@@ -236,15 +236,38 @@
 
   describe 'move_to_dashboard' do
     let(:call) { ApplicationRecord.move_to_dashboard }
-    context 'there are expired records' do
+    context 'there is a expired record' do
```
dfaulken: Whoops, should be `an`.

spec/models/application_record_spec.rb
```diff
@@ -236,15 +236,38 @@
...
+      end
+      it 'calls move_to_dashboard on expired record' do
+        expect_any_instance_of(ApplicationRecord)
+          .to receive(:move_to_dashboard)
+        expect(JobappsMailer).to receive(:saved_application_notification)
+        call
+      end
+    end
+    context 'there are many expired records' do
+      let!(:expired_saved_record1) do
```
dfaulken: Prefer an underscore before the number, e.g. `expired_saved_record_1`.

spec/models/application_record_spec.rb
```diff
@@ -253,10 +276,13 @@
         create :application_record,
                saved_for_later: true,
                date_for_later: Date.tomorrow,
-               note_for_later: 'SO required'
+               note_for_later: 'SO required',
```
dfaulken: 😆

app/models/application_record.rb
dfaulken: Oh, also, it's not strictly necessary to capitalize `NOT NULL`, just FYI.

spec/controllers/application_records_controller_spec.rb
AkivaGreen: Yes, I just realized there is no comma there. I will make sure to double check the tests.

dfaulken: Ah, I was referring to the indentation, actually.

app/models/application_record.rb
AkivaGreen: How about the word `fill` in favor of `build` since that is what the method is really doing.

dfaulken: 👍

app/views/jobapps_mailer/saved_applications_notification.text.erb
AkivaGreen: Applied on would be logical

spec/controllers/application_records_controller_spec.rb
AkivaGreen: Oh, my apologies, yes Rubocop did the indentation.

dfaulken:
@AkivaGreen, do you think you will be able to finish this before the semester ends? If not, please open a branch of this repo, not a fork, and push your changes to it so that we can complete this for you.

dfaulken:
I'll finish this out today.

dfaulken:
Closing in favor of #279.

PR #276, opened 2016-12-06
pre-fill application template email

### LINKED ISSUE #273: Do away with reply-to email
I thought this was already an issue, but I can't find it anywhere.

Since gmail removes the 'reply-to' attribute we set, we should instead add a line in any relevant emails stating that one should not reply to this address, but instead direct questions to application_template.email--and whether that displays should be configurable. Not sure what exactly this should look like, I'm going to tinker with it a bit.

Anbranin:
closes #273

Turns out the params for "email" were just getting sent as blank because the value wasn't pre-filled.

@werebus... I have failed. It was 40 characters.

### DIFF
--- a/app/views/application_drafts/edit.haml
+++ b/app/views/application_drafts/edit.haml
@@ -79,7 +79,9 @@
         .field_attribute_header1
           Reply-to email:
         .field_attribute
-          =f.text_field :email, size: 50
+          =f.text_field :email,
+            value: @draft.application_template.email,
+            size: 50
       .actions
         = f.submit 'Save changes and continue editing'
         = f.submit 'Preview changes'


Anbranin:
Well, pressing the button "save changes and continue editing" calls `update` on the draft. In that method, `@draft.update email: draft_params[:email]`, which is blank if the field doesn't have a value.
The field doesn't have a value because when you publish an application template, the draft gets discarded. The new draft doesn't have an email. It is given the questions in `create_draft` in ApplicationTemplate, a user, and a template, but no other attributes.

dfaulken:
Ah, so then perhaps arguably the problem is that the draft attributes are not the same as the template attributes? Or do I misunderstand?

Anbranin:
Quite. Not all of them. Perhaps a more future-proof thing to do here would be to just assign all the attributes except for the ones we don't want.

dfaulken:
You're the primary contributor, so let's go with whichever approach you feel is best. Feel free to merge this if you'd like.

Anbranin:
I'll merge this so this particlar problem is fixed then look into a refactor.

werebus:
Yup, `create_draft` should probably also populate the `email` field.  A side-effect of this fix is that you can't have a draft that you're in the middle of working on that has a different reply-to than it's template.

Anbranin:
Oooh rite good point

PR #278, opened 2016-12-07
added requested page-view option to datatable

Anbranin:
Don wanted to view all of them at once. Probably when there are like a million past applications this won't be good any more, but it's very, very easy to change.

### DIFF
--- a/app/assets/javascripts/datatables.js.coffee
+++ b/app/assets/javascripts/datatables.js.coffee
@@ -1,4 +1,3 @@
 $(document).ready ->
-  $('.data_table').DataTable({
-    'lengthMenu': [[10, 25, 50, -1], [10, 25, 50, 'All']]
-    });
+  $('.data_table').DataTable lengthMenu: [[10, 25, 50, -1], [10, 25, 50, 'All']]
+return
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/datatables.js.coffee
```diff
@@ -1,2 +1,4 @@
 $(document).ready ->
-  $('.data_table').DataTable()
+  $('.data_table').DataTable({
+    'lengthMenu': [[10, 25, 50, -1], [10, 25, 50, 'All']]
+    });
```
dfaulken: Keep in mind that this is CoffeeScript, hence the CC error.

This could be one line:

```coffee
$('.data_table').DataTable lengthMenu: [[10, 25, 50, -1], [10, 25, 50, 'All']]
```

app/assets/javascripts/datatables.js.coffee
```diff
@@ -1,2 +1,3 @@
 $(document).ready ->
-  $('.data_table').DataTable()
+  $('.data_table').DataTable lengthMenu: [[10, 25, 50, -1], [10, 25, 50, 'All']]
+return
```
werebus: `return`?

PR #280, opened 2017-01-04
we do not need 2 bootstraps

Anbranin:
This fixes nothing but for real we did not need that second one


dfaulken:
> This fixes nothing

Does it also break nothing?

werebus:
FWIW, `twitter-bootstrap-rails`, may do some more things than just JS/CSS.  There are a handful of view helpers in there, for example.

werebus:
*  JQuery UI:  There's [a gem][1] and [a copy in `vendor/assets`][2].
*  Jquery Datatables:  [gem][3], [`vendor/assets`][4]
*  Jquery Datetime Picker:  [gem][5], [`vendor/assets`][6]

I also suspect that a number of the files *in* `vendor/assets` aren't used at all (aren't `require`d by the asset pipeline); we certainly have both the minified and un-minified copies of a lot of them.

Also, also, We have a group called `assets` that only has one of the asset gems in it?

[1]:  https://github.com/umts/jobapps/blob/master/Gemfile#L12
[2]:  https://github.com/umts/jobapps/blob/master/vendor/assets/javascripts/jquery-ui.min.js
[3]:  https://github.com/umts/jobapps/blob/master/Gemfile#L8
[4]:  https://github.com/umts/jobapps/blob/master/vendor/assets/javascripts/jquery.dataTables.js
[5]:  https://github.com/umts/jobapps/blob/master/Gemfile#L9
[6]:  https://github.com/umts/jobapps/blob/master/vendor/assets/javascripts/jquery.datetimepicker.js

dfaulken:
Thanks @werebus — @Anbranin we may as well iron out all of those things while we're here.

Anbranin:
Low priority, will get to it,

Anbranin:
When I added the files to vendor I definitely did not know how gems work.

Anbranin:
> Also, also, We have a group called assets that only has one of the asset gems in it?
We have no assets group--in fact, Rails 4 removed support for it. Am I missing something? THIS IS LIKELY

Anbranin:
There is a chance that what I just did with the datepicker gem will resolve #281 but I'm not sure, and I don't feel like downloading another browser right now? I do not like front-end. :(

Anbranin:
@dfaulken you need to approve. I can override it but I want to make sure all the changes _you_ requested have been completed.

PR #284, opened 2017-02-15
fixed issue with attributes not carrying over

### LINKED ISSUE #277: Give application_drafts the same attributes as their respective template
Change Template#create_draft to give the drafts all the attributes of their template except the id or whatever

artarazavi:
closes #277


Anbranin:
Also, run rubocop..

artarazavi:
Think its good to go now

PR #288, opened 2017-03-09
Upgrade rails

Anbranin:
The hardest part was picking a new name for "ApplicationRecord."
Hope y'all like the name "FiledApplication" because if you don't I got nothin'. I thought about it for a looong time.

Update: "ApplicationSubmission" is the new thang. It's pretty great.


dfaulken:
Since we have `ApplicationDraft` and `ApplicationTemplate`, it seems to me that keeping consistent with having the second word in the model modify "application" would help express that these are related, but distinct, objects. `ApplicationSubmission`, perhaps?

app/controllers/application_drafts_controller.rb
```diff
@@ -34,9 +34,10 @@ def remove_question
 
   def update
     draft_params = params.require(:draft)
-    question_params = draft_params.require(:questions_attributes).permit!
+    question_attributes = draft_params.require(:questions_attributes)
+                                      .permit!.to_h
```
werebus: Can you explain this one? I have a vague feeling that we're doing something questionable here.

spec/controllers/application_templates_controller_spec.rb
```diff
@@ -120,7 +121,7 @@
       end
       it 'redirects back' do
         submit
-        expect(response).to redirect_to :back
+        expect(response).to redirect_to 'http://test.host/redirect'
```
werebus: Interesting. Is this test accurately named now?  It seems like it's redirecting *somewhere*, but is it "back"?

app/controllers/application_drafts_controller.rb
Anbranin: It does mean we're permitting any and all questions_attributes, so I could sanitize a bit to make sure that only the question attributes we expect get permitted. I just copied what was done before in that case, thinking that a staff member screwing with question attributes was unlikely and couldn't cause a security problem.
As for the "to_h" thing, params is no longer a hash in Rails 5. `update_questions` in the draft model expects a hash, not an ActionController::Parameters object, so somewhere it needs to get turned into a hash--here seemed reasonable, given that I don't want a model method to deal with params ever.

spec/controllers/application_templates_controller_spec.rb
Anbranin: On line 108:
`request.env['HTTP_REFERER'] = 'http://test.host/redirect'`
So yes. If you think it would make it clearer, on the next line I could totally
`@back = request.env['HTTP_REFERER']`

Anbranin:
Okay, `ApplicationSubmission` seems great! I'll make those changes. Thanks @dfaulken.

spec/controllers/application_templates_controller_spec.rb
werebus: Nah.  That line was hidden from the diff and I assumed it was some sort of magic-test-value. Looking at the whole test  context it's clear enough.

app/controllers/application_drafts_controller.rb
werebus: Well, when you say it "expects" a `Hash`, in this case you mean that it expects to be able to call `.blank?` and `.each` on it. It's clear that the Rails authors intended a `ActionController::Parameters` object to behave *like* a `Hash`, in that you can access its values with `#[]`, and can use it in `ActiveRecord` assignments. But it doesn't have `.blank?`, for some reason.

My alarm bells we set off because it *seemed* (but wasn't) like a thing you might do to git around a mass-assignment protection or something.

In this case the only vaguely questionable thing here is `.permit!`, but you're right that that was there before.

app/controllers/application_templates_controller.rb
```diff
@@ -27,7 +27,7 @@ def toggle_active
     else show_message :inactive_application,
                       default: 'The application is now inactive.'
     end
-    redirect_to :back
+    redirect_back fallback_location: main_dashboard_path
```
dfaulken: Is there any way to set this more globally? Seems pretty repetitive to write it everywhere.

werebus: My brief Googling says that there isn't a built-in way, no.  You could write a method that did that, I guess.

I think that the thinking was that it'd be pretty rare that the fallback "back" location would be the same site-wide.

dfaulken: You know, actually, you're right. Seems like we shouldn't have the same fallback location everywhere. Here, for example, the main dashboard isn't where we would expect `:back` to normally be.

In general, shouldn't the fallback be wherever we expect they've come from?

dfaulken:
We're still at 'will write tests for this', BTW.

Anbranin:
Yup, just got back to working on this after working on the rodeo app. GETTIN' THERE

Anbranin:
screen shot 2017-05-08 at 10 58 50 pm
LOL WAT

Anbranin:
K ALL SET

PR #289, opened 2017-03-20
number of saved applications now displayed on dashboard

Anbranin:
I did ask the Training Supervisor about this, and he says it's great.

screen shot 2017-03-21 at 3 16 26 pm

screen shot 2017-03-21 at 4 04 26 pm

### DIFF
--- a/app/controllers/positions_controller.rb
+++ b/app/controllers/positions_controller.rb
@@ -20,7 +20,8 @@ class PositionsController < ApplicationController
   end
 
   def saved_applications
-    @saved = @position.saved_applications
+    @saved = ApplicationRecord.where(saved_for_later: true,
+                                     position: @position)
   end
 
   def edit
--- a/app/controllers/positions_controller.rb
+++ b/app/controllers/positions_controller.rb
@@ -20,7 +20,7 @@ class PositionsController < ApplicationController
   end
 
   def saved_applications
-    @saved = ApplicationRecord.where(saved_for_later: true, position: @position)
+    @saved = @position.saved_applications
   end
 
   def edit
--- a/app/controllers/positions_controller.rb
+++ b/app/controllers/positions_controller.rb
@@ -20,8 +20,7 @@ class PositionsController < ApplicationController
   end
 
   def saved_applications
-    @saved = ApplicationRecord.where(saved_for_later: true,
-                                     position: @position)
+    @saved = @position.application_records.where(saved_for_later: true)
   end
 
   def edit
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -39,6 +39,6 @@ class SessionsController < ApplicationController
   end
 
   def new_spire
-    (User.pluck(:spire).map(&:to_i).last + 1).to_s.rjust(8, '0') + '@umass.edu'
+    (User.pluck(:spire).map(&:to_i).max + 1).to_s.rjust(8, '0') + '@umass.edu'
   end
 end
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -39,6 +39,6 @@ class SessionsController < ApplicationController
   end
 
   def new_spire
-    (User.pluck(:spire).map(&:to_i).last + 1).to_s.rjust 8, '0'
+    ((User.pluck(:spire).map(&:to_i).last + 1).to_s.rjust 8, '0') + '@umass.edu'
   end
 end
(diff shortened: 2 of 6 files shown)


app/views/dashboard/_pending_applications.haml
```diff
@@ -10,5 +10,6 @@
               %li= link_to record.user.proper_name, record
           - else
             No pending applications
-        = link_to 'View saved applications for ' + position.name,
+        - count = position.saved_applications.count
```
dfaulken: Should be an instance variable (the applications, at least, if not the count).

app/models/position.rb
```diff
@@ -18,6 +18,10 @@ def name_and_department
     "#{name} (#{department.name})"
   end
 
+  def saved_applications
```
dfaulken:
```ruby
def saved_applications
  application_records.where(saved_for_later: true)
end
```

should be okay, I think.

app/views/dashboard/_pending_applications.haml
```diff
@@ -10,5 +10,6 @@
...
+        = link_to "View saved applications for #{position.name} (#{count})",
```
dfaulken: I maintain that this should be wrapped in a conditional checking for none. If we're worried about UI elements being inconsistent, perhaps some text (`No saved applications`) in the event where there are 0?

dfaulken:
As far as the UI goes, should the link be indented (padding) the same amount as the position name? Looks like it's a little less.

app/views/dashboard/_pending_applications.haml
Anbranin: Here's the thing: The `position` referenced here is something being iterated through in the view. There is currently no way to turn `position.saved_applications` into an instance variable, given that we get it from an instance variable called `@departments`.

app/models/position.rb
Anbranin: woops I forgot how 2 rails

app/views/dashboard/_pending_applications.haml
Anbranin: I agree with having the text "no saved applications." I did not agree with just showing the default text that it has shown for months when there are none. Will make that change.

Anbranin:
That was a deliberate stylistic choice. I liked it.

app/views/dashboard/_pending_applications.haml
dfaulken: We discussed in person just now iterating through in a similar manner to pending applications.

app/views/dashboard/_pending_applications.haml
```diff
@@ -15,4 +15,4 @@
           = link_to "View saved applications for #{position.name} (#{count})",
             saved_applications_position_path(position)
         -else
-          No saved applications
+          %ul No saved applications
```
dfaulken: Does it also look consistent if there *are* saved applications?

Anbranin: YES NOW IT DOES

dfaulken:
Is this ready for review again?

Anbranin:
YES

app/controllers/positions_controller.rb
```diff
@@ -20,7 +20,8 @@ def destroy
   end
 
   def saved_applications
-    @saved = ApplicationRecord.where(saved_for_later: true, position: @position)
+    @saved = ApplicationRecord.where(saved_for_later: true,
```
dfaulken: Or
```ruby
@position.application_records.where(saved_for_later: true)
```

app/controllers/sessions_controller.rb
```diff
@@ -39,6 +39,6 @@ def find_user
   end
 
   def new_spire
-    (User.pluck(:spire).map(&:to_i).last + 1).to_s.rjust 8, '0'
+    ((User.pluck(:spire).map(&:to_i).last + 1).to_s.rjust 8, '0') + '@umass.edu'
```
dfaulken: The parentheses that were originally omitted were around the `rjust`, not the entire expression.

app/views/dashboard/_pending_applications.haml
```diff
@@ -10,5 +10,9 @@
               %li= link_to record.user.proper_name, record
           - else
             No pending applications
-        = link_to 'View saved applications for ' + position.name,
-          saved_applications_position_path(position)
+        - if @saved_records[position].present?
+          - count = @saved_records[position].count
+          %ul= link_to "View saved applications for #{position.name} (#{count})",
```
dfaulken: Since both cases include a `ul`, it'd be better to write it outside of the conditional.

Anbranin:
@dfaulken ready again for review.

PR #294, opened 2017-03-28
Staff interface

### LINKED ISSUE #285: Interface does not have a way to make a current user into staff
If someone has a User object (say, me) in the database who is not staff, there is no current way to make them into staff using the website, you must go into the console. Trying to "create a new staff member" only leads to an invalidation of uniqueness on spire id.

AkivaGreen:
I want to resolve #285 with this pull request.

### DIFF
--- a/app/assets/javascripts/users/promote.js.coffee
+++ b/app/assets/javascripts/users/promote.js.coffee
@@ -1,6 +1,3 @@
 $(document).ready ->
-  $ ->
-    users = $("#promote-search").data("users")
-    $("#promote-search").autocomplete source: users
-    return
+  $('#promote-search').autocomplete source: $('#promote-search').data('users')
   return
--- a/app/assets/javascripts/users/promote.js.coffee
+++ b/app/assets/javascripts/users/promote.js.coffee
@@ -1,0 +1,14 @@
+$(document).ready ->
+  $('.promote-table').on 'change', '#select-all', ->
+    if $(this).is(':checked')
+      $('.promote-select').prop 'checked', true
+    else
+      $('.promote-select').prop 'checked', false
+    return
+  $('.promote-table').on 'change', '.promote-select', ->
+    if $('.promote-select:checked').length == 0
+      $('#select-all').prop 'checked', false
+    else if $('.promote-select:checked').length == $('.promote-table').find('tbody > tr').length
+      $('#select-all').prop 'checked', true
+    return
+  return
--- a/app/assets/javascripts/users/promote.js.coffee
+++ b/app/assets/javascripts/users/promote.js.coffee
@@ -1,14 +1,6 @@
 $(document).ready ->
-  $('.promote-table').on 'change', '#select-all', ->
-    if $(this).is(':checked')
-      $('.promote-select').prop 'checked', true
-    else
-      $('.promote-select').prop 'checked', false
-    return
-  $('.promote-table').on 'change', '.promote-select', ->
-    if $('.promote-select:checked').length == 0
-      $('#select-all').prop 'checked', false
-    else if $('.promote-select:checked').length == $('.promote-table').find('tbody > tr').length
-      $('#select-all').prop 'checked', true
+  $ ->
+    users = $("#promote-search").data("users")
+    $("#promote-search").autocomplete source: users
     return
   return
--- a/app/assets/stylesheets/users.scss
+++ b/app/assets/stylesheets/users.scss
@@ -1,0 +1,14 @@
+.promote-table{
+  border-collapse: collapse;
+  width: 100%;
+  margin: 10px 0px;
+  th, td{
+    border: 1px solid black;
+    height: 50px;
+    text-align: center;
+  }
+}
+
+#promote{
+  margin: 10px 0px;
+}
--- a/app/assets/stylesheets/users.scss
+++ b/app/assets/stylesheets/users.scss
@@ -1,4 +1,3 @@
 #promote-search{
-  margin: 0px 10px 0px 5px;
   width: 40%;
 }
--- a/app/assets/stylesheets/users.scss
+++ b/app/assets/stylesheets/users.scss
@@ -1,14 +1,4 @@
-.promote-table{
-  border-collapse: collapse;
-  width: 100%;
-  margin: 10px 0px;
-  th, td{
-    border: 1px solid black;
-    height: 50px;
-    text-align: center;
-  }
-}
-
-#promote{
-  margin: 10px 0px;
+#promote-search{
+  margin: 0px 10px 0px 5px;
+  width: 40%;
 }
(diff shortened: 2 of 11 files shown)


AkivaGreen:
screen shot 2017-03-28 at 2 46 25 pm

dfaulken:
Can we see what this looks like with a production-ish number of users? `User.count` gives me 590 in production, for example.

Anbranin:
I wasn't really thinking about adding a new page for this, not that it's a bad idea in general. The "Add new staff member" page should be able to take care of this by finding the user by their spire ID and then just giving them an attribute of "staff: true." All you actually needed to do was add a block in the existing controller action.
If you want to keep a page like this, might I suggest figuring out a way around the problem that Dave pointed out -- that there are currently 590 users in the system and that number will grow into the thousands in just a few short years -- and integrating it somehow with the "add new staff" page. One should not have to already know whether a user is part of the system already in order to promote them to staff.

dfaulken:
@Anbranin and I discussed yesterday that this is actually an interesting UI challenge: how to write a nice page that seamlessly integrates "finding an existing thing" with "creating it if required".

app/controllers/users_controller.rb
```diff
@@ -35,21 +35,18 @@ def update
   end
 
   def promote
-    @users = User.where.not(staff: true).sort_by(&:proper_name)
+    users = User.where.not(staff: true)
+    @users = []
+    users.each do |user|
```
dfaulken: You can do this more simply (without maintaining two copies of the same data structure) using map.

AkivaGreen:
screen shot 2017-03-31 at 2 58 01 pm
To make it easier, I made the home page link go to this page, and the here link goes to the create new staff member.

AkivaGreen:
screen shot 2017-03-31 at 2 59 19 pm

AkivaGreen:
All that's left are the tests to get coverage back up to 100%.

app/controllers/users_controller.rb
```diff
@@ -34,6 +34,25 @@ def update
...
+      @users << "#{user.full_name} #{user.spire}"
+    end
+  end
+
+  def promote_save
+    user = User.find_by(spire: params[:user].split.last)
+    if user.nil?
+      redirect_to promote_users_path
+    elsif user.update(staff: true)
```
Anbranin: can we maybe just handle the case that something goes wrong? Perhaps populate flash[:error] or whatever it is?

Anbranin: never miiind

app/assets/javascripts/users/promote.js.coffee
```diff
@@ -0,0 +1,6 @@
+$(document).ready ->
+  $ ->
```
dfaulken: I'm not sure this line (and its accompanying `return`) is necessary. Can we compare this with coffee script in other repos?

app/assets/stylesheets/users.scss
```diff
@@ -0,0 +1,4 @@
+#promote-search{
+  margin: 0px 10px 0px 5px;
```
dfaulken: Why's that? In general, uneven CSS margins/paddings are a flag that there's some other weird problem.

app/controllers/users_controller.rb
```diff
@@ -34,6 +34,25 @@ def update
     end
   end
 
+  def promote
+    users = User.where.not(staff: true)
+    @users = []
+    users.each do |user|
+      @users << "#{user.full_name} #{user.spire}"
+    end
```
dfaulken: wat

This seems inefficient

app/assets/stylesheets/users.scss
AkivaGreen: It doesn't look weird or anything without it. I was just going to liberal with my design. So I can just delete it out since it doesn't really look any different without it.

app/controllers/users_controller.rb
AkivaGreen: I will replace `each` with `find_each` which is a more efficient alternative.

app/assets/javascripts/users/promote.js.coffee
AkivaGreen: I found that the autocomplete would not work without it. [Here](https://jqueryui.com/autocomplete/) you can see the javascript code for a working autocomplete. When converted into coffee script it looks similar to what is shown here.

spec/controllers/users_controller_spec.rb
```diff
@@ -37,10 +37,21 @@
...
-        patch :update, id: user, user: attrs
+        put :update, id: user, user: attrs
         user.reload
         expect(user).to eql initial_user
       end
     end
   end
+  describe 'PUT #promote_save' do
+    context 'promoting a user as staff' do
+      it 'does not promotes the user' do
```
Anbranin: small typo:
"it does not promote"

app/assets/javascripts/users/promote.js.coffee
dfaulken: Right, but you're already inside of a `$(document).ready()` call. You shouldn't then need the `$()` call inside of *that*. That is:

```js
$(document).ready(function(){
  $('#tags').autocomplete({
    //...
  });
});
```

is fine.

app/controllers/users_controller.rb
dfaulken: That's true. I was thinking something more like

```ruby
@users = User.where.not(staff: true)
             .pluck(:first_name, :last_name, :spire)
             .map do |first_name, last_name, spire|
               "#{first_name} #{last_name} (#{spire})"
             end
```

`pluck` is good when you only want some attributes of users, and not to call any methods on them. You can them grab just those data attributes and iterate over them, which has two advantages:

1. Only a single DB query.
2. A much lighter data structure over which to iterate.

app/controllers/users_controller.rb
```diff
@@ -34,6 +34,25 @@ def update
     end
   end
 
+  def promote
+    users = User.where.not(staff: true)
+    @users = []
+    users.find_each do |user|
+      @users << "#{user.full_name} #{user.spire}"
```
Anbranin: To quote @dfaulken:

> That's true. I was thinking something more like

```ruby
@users = User.where.not(staff: true)
             .pluck(:first_name, :last_name, :spire)
             .map do |first_name, last_name, spire|
               "#{first_name} #{last_name} (#{spire})"
             end
```

> `pluck` is good when you only want some attributes from a collection, and when you don't need to call any methods on them. You can them grab just those attributes and iterate over them, which has two advantages:

> 1. Only a single DB query.
> 2. A much lighter data structure over which to iterate.

Anbranin:
@AkivaGreen, since we've set up branch protection on our apps, you no longer need to have a fork. I'd like you to delete your fork and remake this branch directly off of UMTS. I have a feeling that the codeclimate / coverage waiting for status to be reported forever might be connected to the fact that you're forking.

AkivaGreen:
I'll close this and open up a new pull request then.

PR #295, opened 2017-03-28
Use moment to parse and format datetimes

Anbranin:
This solution, though it requires keeping assets in vendor until the `+1` bug is fixed, does eliminate the 1899 issue.

We can also apply it to umts/ft-forms in the same exact manner.

- @dfaulken on @anbranin's computer


Anbranin:
PR authors can't approve their own PRs lol duh

Anbranin:
merged as admin, important that this get done soon.

PR #296, opened 2017-04-27
Codecliamte

dfaulken:
Fixin' some codeclimate issues


Anbranin:
No, you didn't.. looks like you fixed some Codecliamte issues.

app/controllers/departments_controller.rb
```diff
@@ -18,11 +18,9 @@ def destroy
     redirect_to staff_dashboard_path
   end
 
-  def edit
-  end
+  def edit; end
 
-  def new
-  end
+  def new; end
```
Anbranin: ugh

app/controllers/users_controller.rb
```diff
@@ -19,11 +19,9 @@ def destroy
     redirect_to staff_dashboard_path
   end
 
-  def edit
-  end
+  def edit; end
```
Anbranin: but FINE okay it's not a big deal, especially since we won't have to even have these in Rails 5 lol

PR #300, opened 2017-06-16
Specify a path for cron

### LINKED ISSUE #292: Saved applications not moving back to dashboard on date
There's a [cron job](https://github.com/umts/jobapps/blob/master/config/schedule.rb#L1-L3) that's supposed to move the application back to the dashboard but something along the way is failing and those application records are just not moving back.

werebus:
Because ruby is installed in /usr/local on our production servers,
`bundle` wasn't found in the path that cron runs with (even given
`/bin/bash -l`).

I'm pretty confident that this closes #292


config/schedule.rb
```diff
@@ -1,3 +1,5 @@
+env 'PATH', '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin'
```
Anbranin: So my question is: Since this means we'll have to add a line like this to every app we have running on the server, why did you choose not to simply configure cron itself? Because this fix means we'll have to just find a way to tell any future developers "by the way, if you want a cron job to run and your app is deployed on this server you have to set the path manually inside your app" which seems wrong to me.

werebus: I guess, although nearly every other aspect of how the app is deployed on a particular server is also in this repository.

It turns out that "configur[ing] cron itself" is a bit complicated. The path has to be set in the crontab that whenever is using, which is, of course, under the control of whenever (and capistrano). This way, we can be sure that the path and the tasks are always both in the same place.

Neither is great, but the alternative is that someone has to log into production, type `sudo crontab -e` and then add exactly the same line that whenever adds with this fix.

Technically speaking, 😬 I think only the "first" application on the server needs this fix... which is an argument in your favor, I guess?

werebus: Interestingly, [someone else][1] figured this out at some point, although I don't like that fix quite as much.

[1]:  https://github.com/umts/bojangles/commit/0d6edb33b0877d3cb7cfb2d4db8b87da2e9eaf3c

dfaulken: Completely forgot about that.

PR #303, opened 2017-06-27
"Upgraded" capistrano to 3.8

werebus:
Well, actually, it was already updated but without the changes to Capfile and config/deploy.rb needed.


werebus:
Codeclimate rubocop disagreement is fine.

PR #322, opened 2018-10-06
Required

### LINKED ISSUE #314: "Reason for rejection" doesn't have to be required
Requested by Training.

If staff would like to reject applicants for no reason whatsoever, I guess that's fine. If you don't email the applicant, then that field is essentially a departmental note which doesn't seem like a thing that should be required.

artarazavi:
closes #314

### DIFF
--- a/lib/date_and_time_methods.rb
+++ b/lib/date_and_time_methods.rb
@@ -8,7 +8,7 @@ module DateAndTimeMethods
   end
 
   def format_date(date)
-    date.strftime '%A, %B %e, %Y'
+    date.strftime '%A, %B %-d, %Y'
   end
 
   def parse_american_date(date)
--- a/app/controllers/application_submissions_controller.rb
+++ b/app/controllers/application_submissions_controller.rb
@@ -63,7 +63,7 @@ class ApplicationSubmissionsController < ApplicationController
                                   application_submission: @record,
                                   user: @record.user
       Interview.create! interview_parameters
-    else @record.deny_with params.require(:staff_note)
+    else @record.deny_with params[:staff_note]
     end
     show_message :application_review,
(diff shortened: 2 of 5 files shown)


lib/date_and_time_methods.rb
```diff
@@ -8,7 +8,7 @@ def format_date_time(datetime, options = {})
   end
 
   def format_date(date)
-    date.strftime '%A, %B %e, %Y'
+    date.strftime '%A, %B %-d, %Y'
```
dfaulken: I'd be curious to know why we need to change this generic method.

Anbranin: @artarazavi may have wanted to write a better commit message, but it's something @werebus and I agreed on. When she ran the test suite, one of the feature tests failed depending on what day she ran it. There is an expectation that a date displayed on some page will match `record.date_for_later`. It should, but it doesn't, because it's looking for "Monday 6 2018" and instead finds "Monday  6 2018". Do you see the difference? It took me a while.
We brainstormed solutions to this, and determined that it was really unnecessary to display dates with two spaces in them IFF the date has one digit.

dfaulken: Makes sense - thanks!

dfaulken: (Also, GH trimmed your extra space 😛 - I was staring at it for a while before I kept reading.)

dfaulken: FYI, if memory serves I think I've written `%A, %B %e, %Y` in several different repos - may be worth us taking a look through and seeing if we want to change any others.

Anbranin: > GH trimmed your extra space
😆

PR #323, opened 2018-11-02
email validation

### LINKED ISSUE #320: 500 with invalid email
```
A Net::SMTPFatalError occurred in application_submissions#review:

  550 5.1.1 <10/6/1999>: Recipient address rejected: User unknown in local recipient table

  app/models/application_submission.rb:76:in `deny_with'
```

Generally email failures take a while to get back to us, but in cases of the email address, "`10/6/1999`" it bounces immediately. 3 possibilities, of which we could implement any combination:

1.  Rescue the exception, provide feedback to the staff member. That said, I don't know that we provide any way to fix this particular problem.
2.  Validate email address a bit. I know [validating email addresses is a nightmare](https://www.regular-expressions.info/email.html), but... even just checking for an `@` would have caught this one.
3.  Send emails as background jobs. A lot of work (so probably not worth it except as a learning experience), but a arguably "good idea" for an action that can timeout or fail in a longish amount of time.

artarazavi:
closes #320


app/views/application_templates/show.haml
```diff
@@ -50,8 +50,12 @@
               .col-md-6
                 .label= u.label field
               .col-md-6
-                = u.text_field field, size: 50, required: true,
-                  value: @current_user.try(field)
+                - if field == :email
```
werebus: Wouldn't it be simpler to just not put it in the `.each`? I know someone was trying to save some repetition but...

Anbranin:
@artarazavi ⚡️

Anbranin:
⚡️

PR #324, opened 2018-11-05
Gemfile update

Anbranin:
Patches security vulnerabilities in various gems.


werebus:
Travis is upset because your `Gemfile.lock` file contains a reference to `rubocop-rspec`, but it's not in the `Gemfile`.  I suspect you added it to auto-update the factories?  If so, you need to `bundle` again now that you've removed it from your Gemfile.

Anbranin:
You're right, thanks. And auto-updating the factories didn't even work. I had to do it by hand. IT'S FINE

Anbranin:
> codeclimate/coverage Expected — Waiting for status to be reported

I'm pretty sure this is going to say this forever. I'm not sure what to do except remove it and add it back in?

PR #328, opened 2018-12-17
Add seed data and migration version numbers

jforgue:
The amount of data I gave the ApplicationSubmission in the seed may be unnecessary, but I figured it would be good to have a full application.
Open to suggestions.


db/seeds.rb
```diff
@@ -3,6 +3,92 @@
...
+                                                  ["Minimum Job Requirements for **Bus Driver**\r\n\r\n + You must have at least three semesters remaining as a UMass Student\r\n + You must be registered for the current or upcoming semester.\r\n + You must be willing to accept a work assignment of at least 15 hours per week during the semester.\r\n + You must be willing to work a minimum of 180 hours per semester for at least two full semesters after training is complete.\r\n + You must possess a driver's license that is valid in the United States.\r\n + You must be at least 18 years of age, and be able to communicate clearly in English.\r\n + You must be able to pass a NIDA-5 panel pre-employment drug screening and agree to participate in random, unannounced, NIDA-5 panel drug testing throughout your employment.\r\n + You must be willing to sign a legally binding contract that includes (but is not limited to) the above terms.\r\n + You must enjoy driving.\r\n", nil, "explanation", 81],
+                                                  ["I understand and meet the above minimum requirements for entering the UMass Transit bus driver training program.", "Yes", "yes/no", 2],
```
jforgue: I am unsure how to break these lines up without ruining the Markdown formatting .

werebus: I would do this:

```ruby
FactoryBot.create :application_submission,
                  data: [['Bus Application', nil, 'heading', 1],
                         [<<~REQS, nil, 'explanation', 81],
                           Minimum Job Requirements for **Bus Driver**

                           +   You must have at least three semesters remaining
                               as a UMass Student.
                           +   You must be registered for the current or
                               upcoming semester.
                           +   You must be willing to accept a work assignment
                               of at least 15 hours per week during the
                               semester.
                           +   You must be willing to work a minimum of 180
                               hours per semester for at least two full
                               semesters after training is complete.
                           +   You must possess a driver's license that is
                               valid in the United States.
                           +   You must be at least 18 years of age, and be
                               able to communicate clearly in English.
                           +   You must be able to pass a NIDA-5 panel
                               pre-employment drug screening and agree to
                               participate in random, unannounced, NIDA-5 panel
                               drug testing throughout your employment.
                           +   You must be willing to sign a legally binding
                               contract that includes (but is not limited to)
                               the above terms.
                           +   You must enjoy driving.
                         REQS
                         ["Local address (street, apt., etc.)", "255 Governors Dr", "text", 3],
                         #ETC.

Comment (werebus) [reply db/seeds.rb]: You shouldn't need `\r`, and the actual newlines in the heredoc get included in the string.

File: db/seeds.rb [hunk 2]
```diff
@@ -3,6 +3,92 @@
 
 exit if Rails.env.test?
 FactoryBot.create :user, first_name: 'David', last_name: 'Faulkenberry', staff: true, spire: '12345678@umass.edu', email: 'dfaulken@umass.edu', admin: true
+user = FactoryBot.create :user, first_name: 'Justin', last_name: 'Forgue', spire: '13579246@umass.edu', email: 'jforgue@umass.edu'
```
Anbranin: This is a public repository and I'm willing to bet that that's your actual spire ID.

db/seeds.rb
```diff
@@ -3,6 +3,92 @@
...
+department = FactoryBot.create :department, name: 'Bus'
+position = FactoryBot.create :position, department: department, name: 'Operator'
+FactoryBot.create :application_submission, data: [["Bus Application", nil, "heading", 1],
```
Anbranin: I guess this is great but I _hate_ that the actual bus application is the ApplicationTemplate we have on seed. Filling it out for development is a stupid chore because of how many questions there are.

db/seeds.rb
jforgue: That is not my spire ID, I made that one up.

db/seeds.rb
Anbranin: There's no programmatic reason to have all the various questions about licensing or the information about drug testing--the only thing that's necessary is to have an example of all the different question types (date selecting, date-time, text-area, etc).

Anbranin: > some sort of mock application that's not so long

YES. Where the questions aren't all required.

db/seeds.rb
Anbranin: Oh I thought you were being dumb my bad

dfaulken: Apparently you also think he's been around a while... not even *your* SPIRE starts with a 1. 😛

dfaulken:
I'm curious what was wrong with the previous seed data.

Migration version numbers are definitely good, though perhaps could have been a separate PR.

db/seeds.rb
Anbranin: 🖕

jforgue:
I suppose I could have added a new application instead of replacing the other one. The issue with the bus application is that it is too cumbersome for developers to fill out when they want to simulate any behavior regarding it.

werebus:
Eh. I kinda agree with @dfaulken.

I think it _is_ relevant that we have (almost) exactly one use-case for this application, and modeling that might help developers see when they are affecting that use-case with their change.

I'm not very sympathetic to the complaint, "My development database, which I have complete and utter control over, isn't exactly the way I want it." If you're doing a ton of hands-on-mouse testing and need a one-question application, that's pretty easy to fix.

In addition, that complaint doesn't really apply to `ApplicationSubmission`s.

Anbranin:
I have now changed my position and completely agree

werebus:
:pancakes:

dfaulken:
So where does this PR stand? Sounds like on the whole, we don't feel like changing seed data is needed, correct?

So should migration versions be a separate/new PR? Those are definitely good.

Or am I misreading the history here?

jforgue:
@dfaulken I believe you are misreading the history. Originally there was no example `application_submission` created at all in the seeds. I then created a fully filled in `application_submission` based of our bus application. This was then changed to a more generic example before we ultimately decided the bus specific one would be the better option.

Anbranin:
Yeah a filled-in application submission would be great

Anbranin:
accidentally clicked "close" sorry

jforgue:
⚡️

PR #333, opened 2019-02-11
Unreject Button 

### LINKED ISSUE #308: Way to "unreject" applications
Sometimes the hiring manager rejects an application by accident. This is rare. What's less rare is that he or she will reject an application and later the applicant will email him with updated information that makes it possible to interview them (their schedules change etc). In either case, the hiring manager would like to be able to "unreject" an application (move it back to pending).

avacmeyer:
closes #308 
screen shot 2019-02-11 at 11 04 06 am

### DIFF
--- a/app/controllers/application_submissions_controller.rb
+++ b/app/controllers/application_submissions_controller.rb
@@ -102,7 +102,7 @@ class ApplicationSubmissionsController < ApplicationController
     end
   end
 
-  def rejected
+  def reject
     @record.reviewed = false
     @record.move_to_dashboard
     redirect_to staff_dashboard_path
--- a/app/controllers/application_submissions_controller.rb
+++ b/app/controllers/application_submissions_controller.rb
@@ -102,8 +102,7 @@ class ApplicationSubmissionsController < ApplicationController
     end
   end
 
-  def reject
-    @record.reviewed = false
+  def unreject
     @record.move_to_dashboard
     redirect_to staff_dashboard_path
   end
--- a/app/controllers/application_submissions_controller.rb
+++ b/app/controllers/application_submissions_controller.rb
@@ -102,6 +102,12 @@ class ApplicationSubmissionsController < ApplicationController
     end
   end
 
+  def rejected
+    @record.reviewed = false
+    @record.move_to_dashboard
+    redirect_to staff_dashboard_path
+  end
+
   private
 
   def create_user
--- a/app/models/application_submission.rb
+++ b/app/models/application_submission.rb
@@ -168,4 +168,8 @@ class ApplicationSubmission < ApplicationRecord
       JobappsMailer.saved_applications_notification records_by_position, address
     end
   end
+
+  def rejected?
+    !saved_for_later? && reviewed? && !interview.present?
+  end
 end
--- a/app/models/application_submission.rb
+++ b/app/models/application_submission.rb
@@ -93,7 +93,8 @@ class ApplicationSubmission < ApplicationRecord
 
   def move_to_dashboard
     update_attributes(saved_for_later: false,
-                      date_for_later: nil)
+                      date_for_later: nil,
+                      reviewed: false)
   end
 
   def self.move_to_dashboard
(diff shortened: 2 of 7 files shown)


jforgue:
Please fix codeclimate and travis errors.

app/controllers/application_submissions_controller.rb
```diff
@@ -102,6 +102,12 @@ def show
     end
   end
 
+  def rejected
+    @record.reviewed = false
+    @record.move_to_dashboard
```
Anbranin: This method happens to work but that's only a side effect here. I'd like this to be a learning experience instead of me just telling you what to do instead so I'll start by asking:
What does `move_to_dashboard` do for you here?

app/models/application_submission.rb
```diff
@@ -168,4 +168,8 @@ def self.send_notification_emails!(records)
       JobappsMailer.saved_applications_notification records_by_position, address
     end
   end
+
+  def rejected?
+    !saved_for_later? && reviewed? && !interview.present?
```
Anbranin: When an application-submission is saved for later, is it reviewed?

config/routes.rb
```diff
@@ -28,6 +28,7 @@
     end
     member do
       post :review
+      post :rejected
```
Anbranin: It rubs me the wrong way that this action is in the past tense. When you call it you're actively rejecting a submission, not looking for ones that have been previously rejected.

app/controllers/application_submissions_controller.rb
avacmeyer: It redirects you to the dashboard so that you can see the application in the not reviewed section.

config/routes.rb
avacmeyer: ok ill change it

app/controllers/application_submissions_controller.rb
Anbranin: I don't think that's happening there. We'll chat on Slack.

Anbranin: So, `record.reviewed = false` sets the record to unreviewed but doesn't save it in that state. You would need to call `save` before that property you set actually gets persisted. However, calling `record.move_to_dashboard` happens to have an `update_attributes` inside it, which saves the record.
You didn't need `move_to_dashboard` at all, you needed to just call `record.save`. Or, you can change the method `move_to_dashboard` so that it does this as well.

app/models/application_submission.rb
avacmeyer: Nope.

Anbranin: So does this need to say `!saved_for_later` specifically I guess is what I'm going for

avacmeyer: nope.

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 100.0
+    "covered_percent": 99.66
```
Anbranin: Looks like you should write some tests for this.

avacmeyer: yep

app/controllers/application_submissions_controller.rb
jforgue: This doesn't look resolved to me. Maybe you and @Anbranin decided it was resolved outside of github?

avacmeyer: yeah we fixed it.

Anbranin:
@jforgue she did make the code changes, what do you mean specifically?
Also, I'll allow the lack of feature testing for now since we need to get this boat in the sea and the tests are thorough. Besides, I would like to set up system tests at some point anyways, at which point I suspect much will be rewritten.

Anbranin:
I'm going to allow _not_ fixing that codeclimate issue since I think a 150 line model may be a bit too restrictive, unless someone thinks differently?

jforgue:
Okay, I see the desired change now. Isn't a little confusing to have two `move_to_dashboard` methods in the same controller?

Anbranin:
> Isn't a little confusing to have two `move_to_dashboard` methods in the same controller?

Meh

Issue #312, opened 2017-10-25
Access controll fails on `ApplicationSubmission#show` with a nonexistant user


werebus:
`@current_user` will be `nil` if the user isn't in our database:

https://github.com/umts/jobapps/blob/4a1db4fbbc64bcdcc8f79f8602b9b1297484ab18/app/controllers/application_controller.rb#L48-L55

[This line][1] should also include a check for `@current_user.present?`

[1]:  https://github.com/umts/jobapps/blob/4a1db4fbbc64bcdcc8f79f8602b9b1297484ab18/app/controllers/application_submissions_controller.rb#L91-L92

dfaulken:
I would write it as:

```ruby
unless @current_user == @record.user || @current_user.try(:staff?)
  deny_access and return
end
```

dfaulken:
I'd also recommend TDD as the way of verifying the bug's fixed.

werebus:
Closed by #341

PR #341, opened 2019-06-07
Tests written and method updated to fix the bug.

Tanvish:
Closes issue: #312 

I updated the show method as was specified on the issue page. I also wrote a test to check for the issue and it passed.

### DIFF
--- a/spec/controllers/application_submissions_spec.rb
+++ b/spec/controllers/application_submissions_spec.rb
@@ -395,6 +395,16 @@ describe ApplicationSubmissionsController do
         expect(response).not_to render_template 'show'
       end
     end
+    context 'record does not belong in the directory' do
+      before :each do
+        when_current_user_is nil
+      end
+      it 'does not allow access' do
+        submit
+        expect(response).to have_http_status :unauthorized
+        expect(response).not_to render_template 'show'
+      end
+    end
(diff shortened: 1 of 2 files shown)


spec/controllers/application_submissions_spec.rb
```diff
@@ -395,6 +395,16 @@
         expect(response).not_to render_template 'show'
       end
     end
+    context 'record does not belong in the directory' do
```
werebus: This `context` description is a little confusing. The context here is that there is no current user at all. It doesn't really matter what the state of the record is. Maybe `'no current user'` or something to that effect?

Tanvish:
Following up on my previous commit to seeing if there are any more changes required?

Tanvish:
Following up on this pull request.

werebus:
:question: 

Both @Anbranin and I have approved it. Do you have any outstanding questions?

Tanvish:
Oh, no just that I am not authorized to merge this pull request for some reason.

werebus:
Ah yes. Sorry.

Our repository permissions aren't entirely uniform. I think we've gone back and forth on whether trainees should be able to click the merge button.

Tanvish:
Oh, alright.

Issue #337, opened 2019-03-15
Email validation is case-sensitive


werebus:
Introduced by #323, the [regex][ev] for validating email addresses doesn't allow upper-case letters in the domain part (after the `@`). We had a user yesterday try to submit an application with an email address of "`theirname@Umass.edu`".

Domain names should [be treated as case-insensitive][rfc].  The **local** part (before the `@`) is not case-insensitive, but for purposes of _validity_ both types of characters should be (and are currently) allowed.

[ev]: https://github.com/umts/jobapps/blob/a7117f56ae46add815253462a8376bea4bbe11ba/app/models/user.rb#L13
[rfc]: https://tools.ietf.org/html/rfc1035

PR #342, opened 2019-06-07
Updated regex to accept case-insensitive email IDs

Tanvish:
Closes: #337 

The regex code for email validation was updated.


werebus:
Whoops. I steered you wrong a bit. I forgot to mention that you checkout master _first_ before checking out you new branch.

Can you make a PR that only has 9661064 in it? I can walk you through it if you need help.

PR #353, opened 2020-01-03
Add Yarn

Anbranin:
I don't think anything needs to be done to the .travis.yml file. The Travis CI docs state that as long as a file named `yarn.lock` is contained in the home directory, the `yarn` command will automatically be run.
We'll see when I open this PR.


werebus:
>  The Travis CI docs state that as long as a file named yarn.lock is contained in the home directory, the yarn command will automatically be run.

Yeah, I've run across this a few times.  I think that this is the case for Javascript/NodeJS projects.  That is, those which say `language: node_js` in the YAML file.

Anbranin:
I suspected as much. In any case, does everything look OK now?

Anbranin:
Yes. I didn't want to do too many things at once.

PR #355, opened 2020-01-06
Move assets

Anbranin:
Moved all JS packages except for bootstrap (which is a project on its own) from the Gemfile to the package.json. Better to get them straight from the source and remove the middle-man. In this case, no rails or ruby specific code that we needed was contained in any.


package.json
```diff
@@ -1,5 +1,8 @@
 {
   "dependencies": {
-    "jquery-datetimepicker": "^2.5.21"
+    "jquery-datetimepicker": "^2.5.21",
```
werebus: Probably good to explicitly declare jQuery as a dependency:

```
warning " > multiple-select@1.5.2" has unmet peer dependency "jquery@1.9.1 - 3".
 ```

We do _have_ a version of jQuery that matches, but yarn is a little unsure to start out with.

File: Gemfile [hunk 2]
```diff
@@ -26,9 +26,6 @@ gem 'whenever', require: false
 
 source 'https://rails-assets.org' do
   gem 'rails-assets-bootstrap', '~> 3.3'
```
werebus: And finally, this one?

werebus: Here, I'm OK using [the (real, sanctioned) gem](https://github.com/twbs/bootstrap-sass).

app/assets/stylesheets/application.css
```diff
@@ -16,7 +16,7 @@
  *= require_self
```
werebus: I know this isn't a thing you did, but `self` and `tree` should really be at the end of the list, right?

Gemfile
werebus: Missed your PR text.  FWIW, it's only a "whole project" if you want to upgrade to Bootstrap 4. Sticking with Bootstrap 3 should be pretty easy.

Anbranin: The whole reason I'm doing this is to upgrade to bootstrap 4 LOL

werebus: It can wait. Just thought it might be nice to fully remove rails-assets.

package.json
Anbranin: To declare this dependency you just straight up add it to the package.json right? then it knows

Anbranin: it

Gemfile
Anbranin: I've already done it in another "branch that got too large so I'm breaking it up into smaller PRs" so it should be quick.

package.json
werebus: Yup.  Just `yarn add jquery` will get you 3.4.1 (the most recent version)

(Which is, weirdly, included in "1.9.1-3")

Anbranin:
99% sure "jquery-timepicker-rails" is unused.

app/assets/javascripts/application.js
```diff
@@ -15,8 +15,9 @@
 //= require dataTables/jquery.dataTables
```
werebus: This one should be removed.

Anbranin:
I fked up I definitely need a CSS file for those them there datatables

app/assets/stylesheets/application.css
```diff
@@ -10,13 +10,13 @@
  * defined in the other CSS/SCSS files in this directory. It is generally better to create a new
  * file per style scope.
  *
- *= require_tree .
  *= require dataTables/jquery.dataTables
```
werebus: Likewise this one.

We should probably add in the bs3 style for now, right?

Anbranin: Yeah, I found it. I can switch it over when needed.
It's funny--I'm only noticing these mistakes during testing because I still have the gems installed 🙄

Anbranin: Whelp it looks terrible.

Anbranin: Used default styling, is much nicer.

werebus: >  Used default styling, is much nicer.

I suspect you missed something (like perhaps the `table` class?) but also, whatever. We'll worry about that for BS4

Anbranin:
Good point about the `table` class, I completely forgot--but yeah, I'm working on adding bs4 as we speak so whatever

PR #357, opened 2020-01-10
Update bootstrap

Anbranin:
This looks huge but it's all the same kinds of changes. Only stylistic.

Edit: That's a lie. They couldn't all be mere stylistic changes. 

- I noticed that `toggleClass('hidden')` doesn't actually work to hide something from view, or show it. This doesn't actually work in production/master right now. I fixed it "while I was there" and it's one line.. that's the only thing I can think of that arguably doesn't belong in this PR. 
- I had to come up with a whole new way to put a red asterisk on required fields, other than in the view. Did it with Javascript and it's great.
- There was the issue of the datepicker image button no longer showing up in the right place due to the `form-control` class. I had to use "input groups" and glyphs, so ended up downloading a new font to use their free glyphs. It was a whole thing.

So yeah, there's a lot, hope someone wants to take this on.


app/views/application_templates/show.haml
```diff
@@ -1,169 +1,153 @@
...
+        Note that you can click and drag to check or uncheck a range of boxes.
+      .row.form-group
         = fields_for :unavailability do |u|
           =render partial: "application_drafts/unavailabilities_calendar",
                 locals: { form: u }
-
+    .row.form-group
       - if @template.eeo_enabled?
-        .heading Voluntary Equal Employment Opportunity Information
+        %h2.heading Voluntary Equal Employment Opportunity Information
```
werebus: This heading being within a `.form-group` makes it not line up with the other headings on the page:

app/views/dashboard/_past_applications.haml
```diff
@@ -1,17 +1,21 @@
-.dashboard_partial
-  .date_range
-    = form_tag past_applications_application_submissions_path, method: :get do
-      List applications for
-      .note_item
-        = select_tag :department_ids,
+.dashboard_partial.container
```
werebus: `.container` at this level makes this box narrower than the rest of the ones in the dashboard (because of added margin).

app/views/dashboard/_reports.haml
```diff
@@ -1,14 +1,21 @@
-.date_range
-  = form_tag csv_export_application_submissions_path, method: :get do
-    Export records for
-    .note_item
+.container
```
werebus: `.container` at _this_ level makes the _content_ of the box have more margin than the others in the dashboard.

app/views/application_submissions/show.haml
```diff
@@ -1,33 +1,38 @@
...
-        .explanation_item= render_markdown prompt
-    - else
-      .data_row
-        .prompt_item= prompt
-        .response_item= response
+  %p
+    %i This application had previously been saved for later review. Your note:
+  %p
+    = @record.note_for_later
+%table.table.table-bordered
```
werebus: Not sure this should be a table.

At the very least, the width needs to be fixed because the presence of a long "explanation" squishes the answers column:



<sub>Questions in green, answers in red</sub>

app/views/application_drafts/edit.haml
```diff
@@ -1,45 +1,49 @@
...
+              = g.select :data_type, Question::DATA_TYPES, 
+                { include_blank: 'Select a data type...', required: !question.new_record? },
+                { class: 'form-control' }
+            .col-3
+              = g.check_box :required
+        Reply-to email:
+        =f.text_field :email, size: 50, class: 'form-control'
+        .actions
+          = f.submit 'Save changes and continue editing', class: 'btn btn-primary'
+          = f.submit 'Preview changes', class: 'btn btn-primary'
```
werebus: Seems like these should be at the end of the document.

Also, maybe a bit of space on those buttons:

Anbranin:
OK one thing though about select boxes. The multiple-select ones on the dashboard look terrible when any bootstrap styles are applied. I don't know how to explain it better than that, I have to provide a screenshot or you have to see it.

app/views/application_drafts/edit.haml
Anbranin: > Seems like these should be at the end of the document.

no.. I can't

werebus: :heavy_check_mark:

app/assets/stylesheets/application.scss
```diff
@@ -10,15 +10,18 @@
  * defined in the other CSS/SCSS files in this directory. It is generally better to create a new
  * file per style scope.
  *
- *= require bootstrap
  *= require jquery-ui
  *= require jquery-datetimepicker/jquery.datetimepicker.css
  *= require multiple-select/dist/multiple-select.scss
  *= require datatables.net-dt/css/jquery.dataTables.min.css
+ *= require font-awesome/css/font-awesome.min.css
```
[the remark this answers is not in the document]
werebus: Also the JS

app/views/application_submissions/show.haml
```diff
@@ -1,33 +1,38 @@
...
+    %tr
+      - if data_type == 'heading'
+        %td
+          %b=prompt.upcase
+      - elsif data_type == 'explanation'
+        %td
+          %i= render_markdown prompt
+      - else
+        %td=prompt
+      %td=response
```
werebus: One last thing, before this PR, headings and explanations spanned both columns:

```haml
%tr
  - if data_type == 'heading'
    %td{colspan=2}
      %b= prompt.upcase
  -elsif data_type == 'explanation'
    %td{colspan=2}
      %i= render_markdown prompt
  -else
    %td= prompt
    %ed= response
```

Anbranin: you mean `%td{colspan: 2}` but yes

PR #364, opened 2020-02-11
Some nice partials

Anbranin:
Just moved some stuff to partials and used form_with instead. No reason just cleanup.

### DIFF
--- a/app/controllers/application_submissions_controller.rb
+++ b/app/controllers/application_submissions_controller.rb
@@ -55,7 +55,7 @@ class ApplicationSubmissionsController < ApplicationController
   end
 
   def review
-    if params.require(:accepted) == 'true'
+    if params[:application_submission][:accepted] == 'true'
       interview_parameters = params.require(:interview)
                                    .permit(:location, :scheduled)
       interview_parameters.require :location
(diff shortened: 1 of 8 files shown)


spec/controllers/application_submissions_spec.rb
```diff
@@ -341,7 +342,7 @@
       let!(:record) { create :application_submission, saved_for_later: false }
       let(:submit) do
         post :toggle_saved_for_later,
-          params: { id: record.id, commit: 'Save for later', application_submission: { mail_to_applicant: true } }
+             params: { id: record.id, commit: 'Save for later', application_submission: { mail_to_applicant: true } }
```
Tanvish: Is there a reason for the extra spacing here?

Anbranin: yes, rubocop wanted it.

app/controllers/application_submissions_controller.rb
```diff
@@ -55,17 +55,9 @@ def past_applications
   end
 
   def review
-    if params.require(:accepted) == 'true'
-      interview_parameters = params.require(:interview)
-                                   .permit(:location, :scheduled)
-      interview_parameters.require :location
```
Tanvish: So they weren't required here because they were included in .permit?

Anbranin: I'm pretty sure that someone wrote `.require` because they wanted to make sure that the parameters were definitely included.
However, if they're not included, `create!` will fail. So we already have protection against creating an interview with incorrect attributes.

Anbranin:
I have answers

PR #368, opened 2020-02-28
don't hide div now that box is checked automatically

Anbranin:
the "tell them why" box should show on page load, as "notify applicant
of denial" is checked"


app/assets/stylesheets/application_records/show.scss
```diff
@@ -6,6 +6,3 @@
 #notification{
   display: none;
 }
-.rejection-message{
-  display: none;
```
Tanvish: Ah, makes sense now.

PR #379, opened 2020-06-11
Rejection message viewable in application review

lacayodaniel:


app/controllers/sessions_controller.rb
```diff
@@ -42,5 +42,19 @@ def find_user
 
   def new_spire
     (User.pluck(:spire).map(&:to_i).max + 1).to_s.rjust(8, '0') + '@umass.edu'
+
```
BenMelz, line 45: Super nitpicky but this blank line can be removed

app/controllers/sessions_controller.rb
```diff
@@ -42,5 +42,19 @@ def find_user
...
   end
 end
+
+
```
BenMelz, line 49: I think one newline is good enough here

db/schema.rb
```diff
@@ -12,15 +12,15 @@
 
 ActiveRecord::Schema.define(version: 2019_09_26_144109) do
 
-  create_table "application_drafts", options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|
+  create_table "application_drafts", id: :integer, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|
```
BenMelz, line 15: I'm not sure what's going on here, or if it was intentional. If not intentional, are you sure you migrated/seeded your database correctly? To my understanding this file should only be affected by writing new migrations, which there aren't any.

db/seeds.rb
```diff
@@ -1,4 +1,4 @@
-require 'factory_bot_rails'
+rquire 'factory_bot_rails'
```
BenMelz, line 1: I think this may have been an accidental change, and may be linked to the aformentioned schema changes.

app/views/application_submissions/past_applications.haml
```diff
@@ -23,6 +24,11 @@
             = format_date_time record.interview.scheduled
         %td
           = record.staff_note
+        %td 
+          - if record.rejection_message.present?
+            = record.rejection_message
+          - else
+            none
```
BenMelz, line 31: This is matter of personal preference, but I think a blank cell would be preferable to 'none'

BenMelz: In fact, maybe you could dry this up and have just `= record.rejection_message`, depending on whether its stored as nil or '' and how `= nil` behaves with HAML

app/controllers/sessions_controller.rb
Anbranin: Fun fact: the newline at EOF isn't even a real carriage return. It's a hidden character.

app/views/application_submissions/past_applications.haml
Anbranin: HAML doesn't care about nil, see staff_note display above.

db/schema.rb
Anbranin: I think these changes might be OK. He merely ran migrations to get his tables and columns and it just happened. I believe it won't break anything for production.

Anbranin: So @lacayodaniel it looks like it's happening to both of us when we migrate, so.. leave it.

db/seeds.rb
lacayodaniel: Total accident, unrelated

spec/features/viewing_past_applications_table_spec.rb
```diff
@@ -40,6 +40,9 @@
   it 'displays the staff note of the applications' do
     expect(page).to have_text record.staff_note
   end
+  it 'displays the rejection message of the application' do
+    expect(page).to have_text record.rejection_message
```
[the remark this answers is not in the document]
Anbranin: I mention this because if it doesn't, we're testing here that `nil` is showing up on the page, which doesn't test anything.

spec/features/viewing_past_applications_table_spec.rb
```diff
@@ -6,9 +6,17 @@
   let(:start_date) { 1.week.ago.strftime('%m/%d/%Y') }
   # datepicker requires m/d/Y format
   let(:end_date) { 1.week.since.strftime('%m/%d/%Y') }
-  let!(:record) { create :application_submission }
+  let! :record do
+    create :application_submission,
+    rejection_message: "it's not you, it's me",
+    staff_note: "they smelled terrible"
+end
```
Anbranin, line 13: This `end` is MISALIGNED 😱

lacayodaniel: so.. so close

PR #381, opened 2020-06-12
Use new framework defaults

Anbranin:
Use all the new framework defaults except for one: We don't use sqlite3 as a database so the default regarding that is unnecessary.

- `action_controller.default_protection_from_forgery`
Protection from forgery is now included in [ActionController::Base](https://github.com/rails/rails/pull/29742/files#diff-4525a7b40df77f1d9b8287da70913c63R75-R76) instead of in the application controller, depending on the default, which I've set to true in the config.
- `active_support.use_sha1_digests`
The previously used algorithm for generating a secure digest, MD5, is broken (one _can_ generate a collision). Also servers running in FIPS mode will just shut down the Rails process if it tries to generate a digest using MD5. This setting refers only to non-sensitive digests and will invalidate existing caches but that should be fine.
- `active_record.cache_versioning`
Rails 5.2 introduces recyclable cache keys. While our app doesn't frequently need to invalidate its cache because it's not that large (so it wouldn't get filled with stale cache data very frequently) why not remove stale cache versions anyway?
- `action_dispatch.use_authenticated_cookie_encryption` and `action_dispatch.use_authenticated_message_encryption`
`AES-256-GCM` is superior to `AES-256-CBC` as a cipher for encryption. Using the new one for cookies will log everyone out. This would be regrettable if someone were in the middle of filling out an application, but we really should do it at some point. I'm thinking of deploying this early in the morning on a Sunday or Monday when everyone younger than me is still sleeping.


Anbranin:
I think I'm going to use more framework defaults

config/application.rb
```diff
@@ -14,5 +14,11 @@ class Application < Rails::Application
     config.assets.paths << Rails.root.join('node_modules')
     config.filter_parameters += [:password, :secret, :spire, :github]
     config.assets.paths << Rails.root.join('node_modules')
```
werebus, line 16: Lines 14 and 16  are the same

config/application.rb
```diff
@@ -14,5 +14,11 @@ class Application < Rails::Application
...
+    config.action_view.form_with_generates_ids = true
+    config.action_controller.default_protect_from_forgery = true
+    config.active_support.use_sha1_digests = true
+    config.active_record.cache_versioning = true
+    config.action_dispatch.use_authenticated_cookie_encryption = true
+    config.active_support.use_authenticated_message_encryption = true
```
werebus, line 22: Since all of these are the 5.2 defaults, can't you just say

```
config.load_defaults 5.2
```

at the top of the `Jobapps::Application` class?

Anbranin: That's what I was missing. I thought that "defaults" meant "if you don't specifically disable it, it'll work" but I did some mental handwaving to get to the place where merely the existence of a commented-out file disabled anything.

werebus:
Travis failure seems real.

PR #382, opened 2020-06-15
Replaces nil for accurate testing

### LINKED ISSUE #345: A few meaningless specs
Got this message a few times when running the specs:

>  Checking for expected text of nil is confusing and/or pointless since it will always match. Please specify a string or regexp instead.

Indeed, this warning highlighted specs that don't actually do anything. They all amount to

```ruby
expect(page).to have_text something_that_is_actually_nil
```
All pages `have_text nil`. These specs don't actually test for anything at all:

* [x] ["submitting application record student has been authenticated and has a user object application template has been marked as inactive shows text explaining that the application is unavailable"](https://github.com/umts/jobapps/blob/876a455b9045972dbb656e87a6d5e45555e37af2/spec/features/application_record_creation_spec.rb#L29)
* [x] ["submitting application record student has been authenticated but has no user object application template has been marked as inactive shows text explaining that the application is unavailable"](https://github.com/umts/jobapps/blob/876a455b9045972dbb656e87a6d5e45555e37af2/spec/features/application_record_creation_spec.rb#L57)
* [x] ["viewing table of past applications displays any interview notes"](https://github.com/umts/jobapps/blob/876a455b9045972dbb656e87a6d5e45555e37af2/spec/features/viewing_past_applications_table_spec.rb#L57)
* [x] ["viewing table of past applications displays the staff note of the applications"](https://github.com/umts/jobapps/blob/876a455b9045972dbb656e87a6d5e45555e37af2/spec/features/viewing_past_applications_table_spec.rb#L41)

lacayodaniel:
Closes #345

### DIFF
--- a/spec/features/application_record_creation_spec.rb
+++ b/spec/features/application_record_creation_spec.rb
@@ -14,6 +14,7 @@ describe 'submitting application records' do
     before :each do
       when_current_user_is student, integration: true
       visit application_path(application_template)
+      application_template.position.update(not_hiring_text: 'custom text')
     end
     it 'automatically fills in the user fields' do
       expect(find_field('First name').value).to eql student.first_name
@@ -22,7 +23,6 @@ describe 'submitting application records' do
     end
     context 'application template has been marked as inactive' do
       it 'shows text explaining that the application is unavailable' do
-        application_template.position.update(not_hiring_text: 'custom text')
         application_template.update active: false
         visit current_url
         # must reload the page for changes to template to take effect
@@ -37,6 +37,7 @@ describe 'submitting application records' do
       when_current_user_is nil, integration: true
       page.set_rack_session spire: spire
       visit application_path(application_template)
+      application_template.position.update(not_hiring_text: 'custom text')
     end
     it 'creates a user object based on how the user fields are filled in' do
       user_attributes = { first_name: 'John',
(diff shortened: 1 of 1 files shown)


spec/features/application_record_creation_spec.rb
```diff
@@ -22,6 +22,7 @@
     end
     context 'application template has been marked as inactive' do
       it 'shows text explaining that the application is unavailable' do
+        application_template.position.update(not_hiring_text: 'custom text')
         application_template.update active: false
         visit current_url
```
BenMelz, line 27: Changes look good to me as is - I do have one small suggestion

The setup code here seems to not be specific to the it block, so it may make sense to move it to a before :each within the context. This isn't necessary at the moment since you only have a single test for this context and the below one, but good practice to be set up for the future.

lacayodaniel: True. For the sake of good practice, I moved it to the before :each

spec/features/application_record_creation_spec.rb
```diff
@@ -14,6 +14,7 @@
     before :each do
       when_current_user_is student, integration: true
       visit application_path(application_template)
+      application_template.position.update(not_hiring_text: 'custom text')
```
BenMelz, line 17: I think you may have interpreted my recommendation lol
You may already know this, but with rspec you can register a before :each hook for each nested context block you make. This allows you to sort of keep getting more specific with nested contexts.

To do what I recommended earlier you would create a new before :each hook, within the `'application template has been marked as inactive' context` block.

Then you would take these three lines and throw them in that new before :each hook

`application_template.update active: false`
`application_template.position.update(not_hiring_text: 'custom text')`
`visit current_url`

PR #383, opened 2020-06-17
Add barebones resume upload feature

### LINKED ISSUE #361: File uploads
We've used the site for programmer applications a few times, and that often requires an out-of-band "email us a resume and/or cover-letter" process.

Might be nice to have a "file" field type.

BenMelz:
Closes #361 

For an administrator, the feature behaves exactly like the EEO and Unavailabilities features, able to be toggled. A user is prompted to upload a resume at the end of the application when enabled. Each resume is tied dependently to an application submission using ActiveStorage. The resume can be downloaded by viewing the application submission.


app/controllers/application_templates_controller.rb
```diff
@@ -79,6 +79,28 @@ def toggle_unavailability_enabled
     redirect_back fallback_location: back_path
   end
 
+  def toggle_resume_upload_enabled
```
Anbranin, line 82: Lotta repetition in code here. It's fine for now, but it would be worth looking at streamlining it for some later optimization.

Anbranin: (like legit, keep it this way for now, it's too much for this PR)

app/views/application_submissions/show.haml
```diff
@@ -41,6 +41,11 @@
       locals: { unavailability: @record.unavailability }
   %p Grey boxes indicate that applicant is unavailable
 
+-if @record.resume.attached?
+  %strong Resume:
+  =link_to @record.resume.filename, rails_blob_path(@record.resume, disposition: 'attachment')
+  %br
```
Anbranin, line 47: general rule of thumb we have around here: Don't use br tags.

app/views/application_templates/show.haml
```diff
@@ -145,8 +145,15 @@
...
           options_for_select(ApplicationSubmission::GENDER_OPTIONS),
           include_blank: 'Prefer not to answer', class: 'custom-select'
-      - unless @current_user.try(:staff?)
-        .form-group
-          = submit_tag 'Submit application', class: 'btn btn-primary'
+    - if @template.resume_upload_enabled?
+      %h2.heading Resume
+      .explanation
+        Please upload a copy of your resume.
+        - #PLACEHOLDER, IDK WHAT TO REALLY PUT
```
Anbranin, line 152: Wording is fine. However, could we make the upload required? If someone has enabled resume upload the chances are slim that they'd want applications without resumes.

PR #397, opened 2020-12-28
Use the icalendar gem

werebus:
The previous format was technically invalid (line-endings), I kinda prefer putting the burden of RFC parsing onto someone else.


app/models/concerns/routable.rb
```diff
@@ -0,0 +1,17 @@
...
+module Routable
+  extend ActiveSupport::Concern
+
+  included do
+    include Rails.application.routes.url_helpers
+  end
+
+  def default_url_options
+    ActionMailer::Base.default_url_options
+  end
```
Anbranin, line 12: Just curious, why define this method? I just don't see it used.

werebus: It's a method override, and is called by many of the methods in `Rails.application.routes.url_helpers`.

```ruby
ClassA = Class.new.send(:include, Routable)
ClassA.new.url_for(Department.last)
#=>  "http://localhost:3000/departments/2"

ClassB = Class.new.send(:include, Rails.application.routes.url_helpers)
ClassB.new.url_for(Department.last)
# ArgumentError (Missing host to link to! Please provide the :host parameter, set default_url_options[:host], or set :only_path to true)
```

Anbranin (approved):

PR #399, opened 2020-12-31
Upgrade to Rails 6

werebus:
All in all a pretty easy upgrade. The only changes to our application code (rather than config) were:

*  Specify whether string uniqueness validations should be case-sensitive. Default currently is `true`, switching to `false` in Rails 6.1. However, in all cases I decided that `false` was a better choice for us anyways.
* A small deprecation regarding the value of `response.content_type`

Work that remains after this is merged:
*  Configuring the [Zeitwerk autoloader](https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#autoloading). It doesn't get turned on until `config.load_defaults` gets chaged to `'6.0'`.
* Switching to webpacker, maybe?


config/initializers/new_framework_defaults_6_0.rb
```diff
@@ -0,0 +1,45 @@
+# Be sure to restart your server when you modify this file.
+#
+# This file contains migration options to ease your Rails 6.0 upgrade.
+#
+# Once upgraded flip defaults one by one to migrate to the new default.
+#
+# Read the Guide for Upgrading Ruby on Rails for more info on each option.
+
+# Don't force requests from old versions of IE to be UTF-8 encoded.
+Rails.application.config.action_view.default_enforce_utf8 = false
```
werebus, line 10: Breaks IE 5-8 support (IE 9 was released in 2011).

config/initializers/new_framework_defaults_6_0.rb
```diff
@@ -0,0 +1,45 @@
...
+
+# Embed purpose and expiry metadata inside signed and encrypted
+# cookies for increased security.
+#
+# This option is not backwards compatible with earlier Rails versions.
+# It's best enabled when your entire app is migrated and stable on 6.0.
+# Rails.application.config.action_dispatch.use_cookies_with_metadata = true
+
+# Change the return value of `ActionDispatch::Response#content_type` to Content-Type header without modification.
+Rails.application.config.action_dispatch.return_only_media_type_on_content_type = false
```
werebus, line 20: Fixed to one use of this in our app already.

config/initializers/new_framework_defaults_6_0.rb
```diff
@@ -0,0 +1,45 @@
...
+# Return false instead of self when enqueuing is aborted from a callback.
+Rails.application.config.active_job.return_false_on_aborted_enqueue = true
+
+# Send Active Storage analysis and purge jobs to dedicated queues.
+Rails.application.config.active_storage.queues.analysis = :active_storage_analysis
+Rails.application.config.active_storage.queues.purge    = :active_storage_purge
+
+# When assigning to a collection of attachments declared via `has_many_attached`, replace existing
+# attachments instead of appending. Use #attach to add new attachments without replacing existing ones.
+Rails.application.config.active_storage.replace_on_assign_to_many = true
```
werebus, line 31: Figured it was safe to change these now, given that we aren't actually _using_ ActiveJob or ActiveStorage on this app.

Anbranin (approved):

PR #402, opened 2021-01-05
Update friendly_id

werebus:
It was pinned to a version from 2015 for no obvious reason that I can see.


Anbranin:
The reason for pinning it would have been from me reading the directions on how to install the gem, and copying and pasting their instructions:

> Add this line to your application's Gemfile:
> gem 'friendly_id', '~> 5.1.0'

Back then I knew much less than now. Were I to install the gem now I might not constrain the version at all.

Anbranin (approved):
oki doki I guess pinning it to minor is fine

PR #404, opened 2021-01-14
Modified `set_current_user` to use spec type

werebus:
Rather than an argument to determine the spec type. Also modified `when_current_user_is` to no longer take a second options hash argument.

@Tanvish, don't worry I'll take care of any merge conflicts with #403, whichever one gets merged first.


Tanvish (approved):
Looks good to me! And thanks! I will get the other two PRs merged today.

PR #405, opened 2021-01-20
Flip Rails 6 defaults

werebus:
These are the last of the Rails 6 defaults to turn on


Tanvish (approved):
Just gotta merge from master to get the latest changes but looks good to me.

PR #416, opened 2021-02-26
Enable some additional security headers

werebus:
(Links to MDN docs in the commit messages)

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -30,6 +30,10 @@ Rails.application.configure do
     config.cache_store = :null_store
   end
 
+  # Set Same-Site on cookies to "Lax"
+  # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite
+  config.action_dispatch.cookies_same_site_protection = :lax
+
   # Store uploaded files on the local file system (see config/storage.yml for options).
   config.active_storage.service = :local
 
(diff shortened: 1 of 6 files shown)


config/environments/development.rb
```diff
@@ -30,6 +30,10 @@
     config.cache_store = :null_store
   end
 
+  # Set Same-Site on cookies to "Lax"
+  # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite
+  config.action_dispatch.cookies_same_site_protection = :lax
```
Anbranin, line 35: The `strict` setting makes sense for us for production, but why would we want the setting to be different for dev/test?

werebus: "Strict" is _really_ strict. It doesn't send cookies unless the request "comes from" a page of the same origin.

What that meant for me in dev was that, if I opened a new tab and typed in, "localhost:3000", I would always be brought to the dev login whether I was currently logged in or not (because the cookie with my user id didn't get sent).

Also, most of our system tests — well, the ones that run in a "real" browser — that rely on "logging in" via rack-session-access rather than by visiting the login page, entering a username and password, and then clicking links from there, won't work.

Since the only thing we store in the session is the user id, this is pretty much fine in production. If we don't get your user id from cookies, we'll just look for you using the Spire id from Shibboleth instead. So a new direct-navigation will _appear_ to behave the same to the end user. But, in development, it's super annoying. So, I dialed it back a bit.

Anbranin: Ooh okay

Anbranin (approved):

PR #418, opened 2021-03-05
Flip on Rails 6.1 defaults

werebus:


app/views/application_submissions/_save_for_later.html.haml
```diff
@@ -2,7 +2,9 @@
   %h3 Application saved for later
 - else
   %h3 Save application for later review
-= form_with(model: @record, url: { action: :toggle_saved_for_later}) do |form|
+= form_with model: @record,
+  url: { action: :toggle_saved_for_later},
+  local: true do |form|
```
werebus, line 7: This is now the default, but the form is _currently_ broken on `master`.

This change could be omitted. But given that Rails keeps switching back and forth on what they think the default should be, I think its better to just be explicit about what kind of form we want (at least for now).

werebus (commented):
The only remaining default is `Rails.application.config.action_dispatch.cookies_same_site_protection`, which is handled over in #416

Anbranin (approved):

PR #432, opened 2021-06-08
Add Saving Interviews For Later

### LINKED ISSUE #431: Add ability to save interviews for later.
Email from Don:

> If possible, what I would like is the ability to take an application that has had an interview, and have the option to move it from the Dashboard and save it for later without accepting, rejecting or hiring.

So, make sure the "save for later" button shows up on the interview page. This is for the case where someone has been accepted to the position after having an interview, but they won't start working or be officially "hired" until much later, later enough that you want it off the dashboard where it's cluttering things.

jpoleary:
Adds ability for staff to save interviews for later, temporarily removing these interviews from the dashboard to reduce clutter. Closes #431

### DIFF
--- a/app/controllers/application_submissions_controller.rb
+++ b/app/controllers/application_submissions_controller.rb
@@ -62,15 +62,15 @@ class ApplicationSubmissionsController < ApplicationController
   end
 
   def toggle_saved_for_later
-   @interview = @record.interview
-   if @interview.present?
-    flash[:message] = 'Interview successfully updated' if @interview.update save_for_later_params
-   elsif @record.update save_for_later_params
-    flash[:message] = 'Application successfully updated'
-   else
-    flash[:errors] = @record.errors.full_messages
-   end
-   redirect_to staff_dashboard_path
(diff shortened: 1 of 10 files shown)


jpoleary:
I have a sneaky feeling that I missed something, but while playing around with the site in localhost I couldn't find any issues that I may have goofed on or missed

Anbranin (changes requested):
I'm sorry that this was misunderstood, I realize now I wasn't that clear on the issue. There was no need to add another column to `Interview`. I think this will be harder to maintain in the future and adds a lot of cyclomatic complexity to a lot of methods.

The things that needed to have happened were:
1) Make the "save for later" button appear on the
 
2) Change the scope `pending` in interviews to no longer return interviews whose application submissions are saved_for_later.

I will say that it's not a _bad_ idea to have a page of "interviews saved for later", but the request was very specific and this was not part of it. I think you've brought up the excellent point that there should be a way to tell which applications saved for later have had interviews. Why not add an extra column "interview date" to the saved-for-later page?

PR #433, opened 2021-06-10
Correctly addresses issue for saving interviews

### LINKED ISSUE #431: Add ability to save interviews for later.

jpoleary:
Closes #431


.ruby-version
```diff
@@ -1 +1 @@
-2.7.3
+2.7.2
```
benmelz, line 1: ?

Gemfile
```diff
@@ -10,7 +10,7 @@ gem 'haml'
 gem 'haml-rails'
 gem 'icalendar'
 gem 'jquery-rails'
-gem 'jquery-ui-rails'
+gem 'jquery-ui-rails', '~>5.0.5'
```
benmelz, line 13: Unless someone told you to do this, this seems like something to add in a separate PR.

Gemfile.lock
```diff
@@ -407,7 +407,7 @@ DEPENDENCIES
   whenever
 
 RUBY VERSION
-   ruby 2.7.3p183
+   ruby 2.7.2p137
 
 BUNDLED WITH
-   2.2.16
+   2.1.4
```
benmelz, line 413: Unless you're adding/explicitly updating gem versions, I don't think this file should have been changed - I'm guessing it's from the ruby downgrade you included?

app/models/interview.rb
```diff
@@ -20,7 +20,7 @@ class Interview < ApplicationRecord
   after_update :resend_confirmation
 
   default_scope { order :scheduled }
-  scope :pending, -> { where completed: false }
+  scope :pending, -> { joins(:application_submission).where(interviews: {completed: false}, application_submission: {saved_for_later: false}) }
```
benmelz, line 23: Spaces in `{}` are missing (rubocop?).

app/models/interview.rb
```diff
@@ -54,7 +54,11 @@ def information(options = {})
   end
 
   def pending?
-    !completed
+    !completed && !saved_for_later
+  end
+
+  def saved_for_later?
+    extract_associated(:application_submission).saved_for_later?
```
benmelz, line 61: We only have one associated `:application_submission` (right?), so using `extract_associated` provides no performance benefits (the preload might actually make this slower- my `ActiveRecord` knowledge is a bit fuzzy right now).

the simpler `delegate :saved_for_later?, to: :application_submission` in the header would suffice.

public/assets/.sprockets-manifest-4f5db62d128c43c9bb0ba2d715df1fa5.json
```diff
@@ -0,0 +1 @@
...:13-04:00","size":4710,"digest":"381ee179a3be9bfee4c78b543207575e2f8267fa910a2a29d9c1dfd936604341","integrity":"sha256-OB7heaO+m/7kx4tUMgdXXi+CZ/qRCiop2cHf2TZgQ0E="},"trans-header-large-f32bf6d302aeb1541f9e62d1e1309f4b6b20ec1eaf05e24a8a63c7b37bd134b1.png":{"logical_path":"trans-header-large.png","mtime":"2021-06-03T15:36:13-04:00","size":129004,"digest":"98a1f8fdc4da32cc142910bdfa1197f11cb53647a13ca1f44a31eb7420e56944","integrity":"sha256-mKH4/cTaMswUKRC9+hGX8Ry1NkehPKH0SjHrdCDlaUQ="},"wordmark-5a854e85808deea210825adb8846ce075251470593029f4e1c28f4a5c0fa42ef.svg":{"logical_path":"wordmark.svg","mtime":"2021-06-03T15:36:13-04:00","size":6555,"digest":"2815e9da48f6243e31bc64959fee770bd4348c49294c02eefe90a2e51034c785","integrity":"sha256-KBXp2kj2JD4xvGSVn+53C9Q0jEkpTALu/pCi5RA0x4U="}},"assets":{"manifest.js":"manifest-dad05bf766af0fe3d79dd746db3c1361c0583026cdf35d6a2921bccaea835331.js","favicon.ico":"favicon-5165f4a6f467d4decc14352b9cd3a2be0e1010144f5bc0d611583f9dceb9e4d6.ico","trans-header-large.png":"trans-header-large-f32bf6d302aeb1541f9e62d1e1309f4b6b20ec1eaf05e24a8a63c7b37bd134b1.png","wordmark.svg":"wordmark-5a854e85808deea210825adb8846ce075251470593029f4e1c28f4a5c0fa42ef.svg"}}
```
benmelz, line 1: I'm fairly certain that all of these `public/assets` files are asset hashes from the compilation options you switched on (see above) that shouldn't be here.

benmelz (changes requested):
I don't have this repo set up currently, but here's visual sweep of things for the moment

config/environments/development.rb
```diff
@@ -65,10 +65,11 @@
   # Debug mode disables concatenation and preprocessing of assets.
   # This option may cause significant delays in view rendering with a large
   # number of complex assets.
-  config.assets.debug = true
+  config.assets.debug = false
 
   # Suppress logger output for asset requests.
   config.assets.quiet = true
+  config.assets.check_precompiled_asset = false
```
benmelz, line 72: Sorry I guess I didn't save my comment in the previous review, but you shouldn't change these options here. (There are special cases where you may want to, but even then you should always switch them back). We pretty much always want debug on in development so that JS/CSS is easier to debug.

app/views/application_submissions/show.haml
```diff
@@ -96,6 +96,7 @@
       .form-group
         Interview occurred on
         = @interview.information
+    = render 'save_for_later', locals: { record: @record }
```
Anbranin, line 99: A little nitpick is that this already exists elsewhere. You can, I think, just change the logic that makes it display.

Anbranin (changes requested):

Anbranin:
You have a codeclimate issue to fix and then this looks gtg. Please run rspec.

Anbranin (changes requested):
There shouldn't be any schema changes included in this PR unless you have a good reason. Careful about your `git add` commands.

Anbranin:
By the way, you don't have to wait until Codeclimate runs to check if you "made it happy". You can just run `rubocop` on your file like so: `$rubocop app/models/whatever.rb`

app/models/interview.rb
```diff
@@ -20,7 +20,12 @@ class Interview < ApplicationRecord
   after_update :resend_confirmation
 
   default_scope { order :scheduled }
-  scope :pending, -> { where completed: false }
+  scope :pending, lambda {
+                    joins(:application_submission).where(interviews: { completed: false },
+                                                         application_submission: { saved_for_later: false })
+                  }
+
+  delegate :saved_for_later?, to: :application_submission
```
Anbranin, line 28: Do you ever actually call `@interview.saved_for_later?` anywhere? Like why have this delegation, I wonder?

jpoleary: Must have forgotten to take it out, will remove

db/schema.rb
```diff
@@ -30,6 +30,7 @@
     t.bigint "byte_size", null: false
     t.string "checksum", null: false
     t.datetime "created_at", null: false
+    t.string "service_name", null: false
```
Anbranin, line 33: what is thiiiiiis

db/schema.rb
```diff
@@ -161,4 +162,5 @@
   end
 
   add_foreign_key "active_storage_attachments", "active_storage_blobs", column: "blob_id"
+  add_foreign_key "active_storage_variant_records", "active_storage_blobs", column: "blob_id"
```
Anbranin, line 165: and thiiiiiis

Anbranin (changes requested):

db/schema.rb
werebus: https://github.com/umts/jobapps/blob/master/db/migrate/20210224185746_add_service_name_to_active_storage_blobs.active_storage.rb

db/schema.rb
werebus: https://github.com/umts/jobapps/blob/1a8c33b31c99e6950668d840767276ee3000fdfd/db/migrate/20210224185747_create_active_storage_variant_records.active_storage.rb#L9

werebus: Accidentally removed in #417 due to a merge error (07a9bb316497c45f90eabbc5ee0ebf754d96acaa)

PR #460, opened 2022-10-25
Fix failing move-to-dashboard cron job

### LINKED ISSUE #292: Saved applications not moving back to dashboard on date

werebus:
Sure has been 5 years off and on that this has been broken...

Closes #292

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -1,4 +1,5 @@
 env 'PATH', '/opt/ruby/bin:/usr/sbin:/usr/bin:/sbin:/bin'
+env 'SECRET_KEY_BASE', 'DOESNOTMATTER'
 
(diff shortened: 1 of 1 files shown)


benmelz (approved):
I'll approve in the name of getting it running, but this is just a bandaid no? Console environment also refuses to boot because of the underlying issue here.

We should probably configure the secret key in `secrets.yml`, or even better, `credentials.yml.enc` as opposed to (I'm guessing here) an environment variable configured by whatever service/daemon setup we have in prod.

If that all tracks, perhaps opening an issue would be wise

PR #463, opened 2022-11-01
Credentials

dirschn:
An attempt to migrate to the Rails built in credentials system.

The production key has already been added to the server at `/srv/jobapps/shared/config/credentials/production.key` and has been added to the Keepass as well under `rails/jobapps/production.key`


config/deploy.rb
```diff
@@ -11,7 +11,8 @@
 
 append :linked_files,
   'config/application.yml',
-  'config/database.yml'
+  'config/database.yml',
+  'config/credentials/production.yml.enc'
```
benmelz, line 15:
```suggestion
  'config/credentials/production.key'
```

Double checking cause of this discrepancy -

> The production key has already been added to the server at /srv/jobapps/shared/config/credentials/production.key

This is definitely 100% correct?

benmelz (changes requested):

config/deploy.rb
dirschn: just looked it up and indeed I misunderstood

benmelz (approved):
Looks good

PR #488, opened 2023-11-09
EEO update

### LINKED ISSUE #485: Change EEO Ethnicity optons
Training coordinator wrote to request the following

1. Change `ApplicationSubmission::ETHNICITY_OPTIONS` to
    * "White (Not of Hispanic origin)"
    * "American Indian/Alaska Native"
    * "Black or African American (Not of Hispanic origin)"
    * "Hispanic or Latino"
    * "Asian American"
    * "Native Hawaiian and Other Pacific Islander"
    * "Multiracial"
2. For those options that are, essentially small language changes, we can go back and change previous selections:
    * "American Indian or Alaskan Native" -> "American Indian/Alaska Native"
    * "Black (Not of Hispanic origin)" -> "Black or African American (Not of Hispanic origin)"
    * "Hispanic" -> "Hispanic or Latino"
    * "Mixed ethnicity" -> "Multiracial"
3. Leave "Asian or Pacific Islander" unchanged in the database; there's no way to know which one those people would have picked.
4. The order in #&#x200B;1 should be the order they're shown on the "EEO Data" page
5. This change should go into effect as close to 2024-01-01 as possible.

---

Note that because this isn't an enum, but rather a text field, some production values won't be among the choices after the change. That's fine, they'll show up as extra data when he does his reports.

dirschn:
closes #485 

I wasn't sure what the "EEO Data" page is, and I also wasn't sure if I should be updating the info in the applications `data` column.. that seemed yucky so I'm holding out hope I don't have to.


app/models/application_submission.rb
```diff
@@ -43,11 +43,12 @@ class ApplicationSubmission < ApplicationRecord
   scope :with_ethnicity, -> { where.not ethnicity: [nil, ''] }
 
   ETHNICITY_OPTIONS = ['White (Not of Hispanic origin)',
-                       'Black (Not of Hispanic origin)',
-                       'Hispanic',
-                       'Asian or Pacific Islander',
                        'American Indian or Alaskan Native',
```
werebus, line 46: You changed the "or" to a "/" in the rake task, but not here.

werebus (changes requested):
["EEO Data page"](https://github.com/umts/jobapps/blob/fdd5e1371500ba5afc84b903b63eb16c5934ed9d/app/controllers/application_submissions_controller.rb#L36-L42).

The order for the two tables that have ethnicity data in them is established [here](https://github.com/umts/jobapps/blob/fdd5e1371500ba5afc84b903b63eb16c5934ed9d/app/models/application_submission.rb#L160) and [here](https://github.com/umts/jobapps/blob/fdd5e1371500ba5afc84b903b63eb16c5934ed9d/app/models/application_submission.rb#L121). So, as long as the `Array` constant is in the order requested it should be fine.

lib/tasks/eeo_update.rake
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+namespace :application_submission do
+  desc 'Update old EEO ethnicities to the new wording'
+  task update_ethnicity: :environment do
+    ApplicationSubmission.each do |app|
```
werebus, line 6: It's only ~3500 records, but...

```suggestion
    ApplicationSubmission.find_each do |app|
```

lib/tasks/eeo_update.rake
```diff
@@ -0,0 +1,21 @@
...
+      case app.ethnicity
+      when 'Black (Not of Hispanic origin)'
+        app.update ethnicity: 'Black or African American (Not of Hispanic origin)'
+      when 'American Indian or Alaskan Native'
+        app.update ethnicity: 'American Indian/Alaska Native'
+      when 'Hispanic'
+        app.update ethnicity: 'Hispanic or Latino'
+      when 'Mixed ethnicity'
+        app.update ethnicity: 'Multiracial'
```
werebus, line 15: Think I'd like to see `update!` here (or some other failure tracking).

werebus (changes requested):

werebus (approved):
I think this is all set, but a reminder that Don wants to wait until 1/1 to merge and deploy.

werebus:
Task has been run in prod.

PR #502, opened 2023-12-27
Use Selenium for webdriver management

werebus:
Selenium 4.11+ knows how to manage chromedriver.

The `webdrivers` gem isn't needed anymore. See umts/screaming-dinosaur#290


frothedoatmilk (approved):
Looks good!

Issue #484, opened 2023-10-06
Remove datetimepicker


werebus:
We've been moving away from datepickers, it would be nice to drop this dependency

Also, datetimepicker is the only reason this app requires moment.js as well.

PR #517, opened 2023-12-29
Replace UI datepickers

werebus:
There are still two more in applications themselves; that, I want to look at a bit more carefully.

First step on #484


app/controllers/application_submissions_controller.rb
```diff
@@ -120,17 +120,13 @@ def record_params
...
-    parameters[:saved_for_later] = params[:commit] == 'Save for later'
-    parameters[:mail_note_for_later] = parameters[:mail_note_for_later] == '1'
-    if parameters[:date_for_later].present?
-      parameters[:date_for_later] = Date.strptime(
-        parameters[:date_for_later], '%m/%d/%Y'
-      )
+    ).tap do |p|
+      p[:saved_for_later] = params[:commit] == 'Save for later'
+      p[:mail_note_for_later] = p[:mail_note_for_later] == '1'
+      p[:date_for_later] = Date.parse p[:date_for_later] if p[:date_for_later].present?
```
benmelz, line 128:
```suggestion
```
Bare column, so AR should handle casting

benmelz (changes requested):

app/controllers/application_submissions_controller.rb
```diff
@@ -23,16 +23,16 @@ def create
 
   def csv_export
     respond_to :csv
-    start_date = parse_date_picker_param(:start_date)
-    end_date = parse_date_picker_param(:end_date)
+    start_date = Date.parse params[:start_date]
+    end_date = Date.parse params[:end_date]
     @records = ApplicationSubmission.in_department(given_or_all_department_ids)
                                     .between(start_date, end_date)
```
benmelz, line 29: Not requesting changes be made here, but eventually I think it would be nice to move any parsing to `.between`, keeping in line with "cast in the model layer".

Also goes for the other class methods below.

werebus: :+1: `.eeo_data` actually just passes those dates to `.between`, so pretty easy.

benmelz: Yeah `&.to_date` should be pretty unobtrusive as well

benmelz (approved):

PR #544, opened 2024-01-19
Revert "Remove datepickers on the actual job applications"

werebus:
Turns out there's all kinds of "dates" in production that are neither an American date nor an ISO8601 date nor an empty string.

I have a fix in mind, but it may take a little bit, and I want the app to be deployable in the meantime.


dirschn (approved):

werebus:
TIL: when you revert a revert commit, git now titles that:

>Reapply "Whatever"

rather than:

>Revert "Revert "Whatever""

PR #546, opened 2024-01-22
Remove datetime pickers

### LINKED ISSUE #484: Remove datetimepicker

werebus:
Closes #484


dirschn:
> frothedoatmilk says...

https://github.com/umts/jobapps/pull/545 run the rake task in this PR

frothedoatmilk:
We shouldn't need to run a rake task after resetting the DB, it should just work.™ Some light substitution was all that was needed #550

werebus:
Seeds are fixed in another PR (#547) — the same PR that removes the rake task that also fixes your problem.

frothedoatmilk (approved):
In that case

werebus:
> One last thing: https://github.com/umts/jobapps/blob/main/spec/system/viewing_past_applications_table_spec.rb#L7 Not sure if you want to change this spec? It still works but the formatting of the input is definitely not necessary

Well, at the very least the comment is wrong (`:db` is Y-m-d).

frothedoatmilk:
Then it should probably be removed, since the specs will run without it now

frothedoatmilk:
I meant also the `to_formatted_s`, no reason to call it and could be confusing later

frothedoatmilk (approved):

PR #552, opened 2024-01-23
Upgrade Rails to 7.0

werebus:


werebus:
Codeclimate issues addressed in #553

werebus:
https://github.com/umts/jobapps/pull/552/files/1d51b7eb4442c07de3c01bd505e8e1babf6503fd to see the changes without the already-reviewed ones in #553

dirschn (commented):
What's up with the switch to double quotes?

dirschn:
Oh, looks like that's from the app:update task?

dirschn (approved):

werebus:
> Oh, looks like that's from the app:update task?

Yeah, at one point "someone" switched them all to single-quotes. But, that someone now thinks that it makes more sense not to have meaningless differences between Rails' default version and ours because then _future_ updates will be easier (`git diff` will show actual differences between our config and the default)

PR #559, opened 2024-01-30
Add step attributes to datetime fields

benmelz:
This is in response to a Don feedback - he is displeased that the native datepickers offer granular (down to the minute) time values and would prefer them to only offer multiples of 15.

Now this doesn't actually do that - it only affects the intervals that the arrow keys change the time. But I'd like to do something to say "we tried" before writing the "sorrrrrry but no" response.

### DIFF
--- a/.haml-lint.yml
+++ b/.haml-lint.yml
@@ -1,4 +1,6 @@
 linters:
+  LineLength:
+    max: 120
   RuboCop:
     ignored_cops:
       - Metrics/BlockNesting
--- a/app/views/application_submissions/_approve.html.haml
+++ b/app/views/application_submissions/_approve.html.haml
@@ -5,9 +5,7 @@
     .row
       .form-group.col
         = f.label :scheduled, 'Date/time'
-        = f.datetime_field :scheduled,
-          class: 'form-control',
-          size: 40, required: true
+        = f.datetime_field :scheduled, required: true, step: 60 * 15, class: 'form-control'
       .form-group.col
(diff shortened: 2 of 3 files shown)


.haml-lint.yml
```diff
@@ -1,4 +1,6 @@
 linters:
+  LineLength:
+    max: 120
```
dirschn, line 3: cheekyyyyyy

dirschn (approved):

werebus:
FWIW, Firefox doesn't provide any sort of time-selection help. Nor does the step appear to affect what I can type or what the arrow keys do. I do get yelled at if I type "11:07" and hit "submit", though, I guess.

benmelz:
> FWIW, Firefox doesn't provide any sort of time-selection help. Nor does the step appear to affect what I can type or what the arrow keys do. I do get yelled at if I type "11:07" and hit "submit", though, I guess.

Ah yes I should've mention this is chrome-centric. Safari also doesn't provide anything either

werebus:
Just for the sake of idle curiosity:

* Mobile Firefox (Android): the UI is a calendar stacked on top of an analog-clock-chooser-thing. It is not constrained by the step, and even worse, there's no client-side validation: the form just doesn't submit.
* Mobile Chrome (Android): The UI is 5 "wheel" selectors, the minutes are unaffected by the step, and you get a client-side validation similar to desktop Firefox, but it at least localizes the two closest times for you.

Issue #610, opened 2024-06-13
Remove or demote Codeclimate


werebus:
I'm getting a bit weary of their rubocop system that requires them to publish their own Rubocop "engine" every time there's an update. They also need to think that the same Rubocop extensions are "important" as we do.

We can keep it, maybe, but let's move linting to someplace else in the CI

benmelz:
I don't really care for it so I vote remove - it shouldn't be an _insurmountable_ task to get our smaller apps into compliance and just all-or-nothing global lint as a ci step

frothedoatmilk:
I don't really see why we're running rubocop on CodeClimate, at the very least. We have Travis doing the same thing, except we have much more control over how Travis works? The other checks can't hurt, although I assume we're paying for code climate...

benmelz:
> I don't really see why we're running rubocop on CodeClimate, at the very least.

I wasn't around when it was added, but it offers an enticing set of features, namely having all your linters run and their outputs displayed prettily in one place, the ability to only enforce new offenses (more on this later), coverage drop checking and a bunch of other nebulous metrics about your code.

> We have Travis doing the same thing, except we have much more control over how Travis works? The other checks can't hurt

Travis doesn't currently run rubocop no, it strictly does the test suite. To get travis to do this and be helpful we either have to get the whole app linted as I mentioned above, or [find some way to check if we're adding _new_ offenses](https://github.com/umts/round-three/pull/6441)

>  although I assume we're paying for code climate...

We're most definitely on a free, open-source plan

frothedoatmilk:
Throughout writing that whole message, I just realized that I mixed up rubocop and rspec. It's definitely the end of the day

werebus:
#656

PR #611, opened 2024-06-13
Use a "base" config for Rubocop

werebus:
This addreses #610 in that Codeclimate won't _fail_ while we work on something more drastic.


werebus:
Of course when I said it _won't_ fail, I meant...

werebus:
Ugh

PR #615, opened 2024-06-18
Don't run rubocop at all in CodeCimate

werebus:
Part of #610, let's try to avoid those red Xs for now


werebus:
Well, not yet... I think **that** red X is on a merge? I also un-required the check.

benmelz (approved):

PR #642, opened 2024-09-10
change wording on dashboard

### LINKED ISSUE #627: Adjust "reviewed" wording on student dashboard
> Good Morning,
>  
> Would it be possible to make it so when an applicant completes the application and submits it, they get a confirmation on the site and are then redirected to anywhere else than the application page?  People seem to be thinking that the app failed, and they resubmit an application.
>  
> Thanks!
>  
> Don

I mean, we do send you back to the "dashboard" and there is a success message, which I'd think would be enough. Guess not though. It would be relatively inexpensive/better to send them directly to the show page application they submitted?

Anbranin:
Closes #627 
Screenshot 2024-09-10 at 9 25 25 AM


benmelz (approved):

app/views/dashboard/student.haml
```diff
@@ -8,32 +8,26 @@
...
-      - application = @application_submissions[position].last
-      - should_show = application.pending? || should_show_denied_applications?
-    - if app_submitted && should_show
-      %li= link_to "Review your application for #{position.name}", application
-    - if position.application_template.try(:active?)
-      - if app_submitted
-        - if should_allow_resubmission?
-          -if application.reviewed
+    // most recent application, if more than one
+    - if application = @application_submissions[position].try(:last)
```
werebus, line 18: `@application_submissions` can be `nil` if the person hasn't applied yet (no `@current_user`).

Also, Rubocop would like you to put `()`s around this assignment to signal that you totally didn't mean `==`

werebus (changes requested):

app/views/dashboard/student.haml
```diff
@@ -8,32 +8,26 @@
...
-      - application = @application_submissions[position].last
-      - should_show = application.pending? || should_show_denied_applications?
-    - if app_submitted && should_show
-      %li= link_to "Review your application for #{position.name}", application
-    - if position.application_template.try(:active?)
-      - if app_submitted
-        - if should_allow_resubmission?
-          -if application.reviewed
+    // most recent application, if more than one
+    - if application = @application_submissions.try(:[], position)&.last
```
werebus, line 18: Two things:

1. Rubocop would suggest (and I agree) that there should be parentheses here to indicate that you intended to do an assignment inside a conditional
2. There are a few ways to do this without `try`:

    ```ruby
    @application_submissions&.dig(position)&.last
    @application_submissions&.fetch(position, nil)&.last
    @application_submissions&.[](position)&.last #Look at majestic string of punctuation
    ```

app/views/dashboard/student.haml
```diff
@@ -8,32 +8,26 @@
...
-      -else
-        %li= link_to "Submit application for #{position.name}",
-          application_path(position.application_template)
     - else
-      %li
-        = position.not_hiring
+      - if position.application_template.try(:active?)
+        %li= link_to "Submit application for #{position.name}", application_path(position.application_template)
+      - else
+        %li= position.not_hiring
```
werebus, line 33: This could be collapsed one `if`-level:

```haml
    - elsif position.application_template.try(:active?)
      %li= link_to "Submit application for #{position.name}", application_path(position.application_template)
    - else
      %li= position.not_hiring
```

werebus (changes requested):

app/views/dashboard/student.haml
```diff
@@ -8,32 +8,28 @@
...
+          %li You are not allowed to resubmit this application.
+      - elsif position.application_template.try(:active?)
         %li= link_to "Submit application for #{position.name}",
-          application_path(position.application_template)
-    - else
-      %li
-        = position.not_hiring
+                      application_path(position.application_template)
+      - else
+        %li= position.not_hiring
```
werebus, line 35: Test failures are correct, this section is indented one level too deep.

As it stands, this won't let anyone apply unless they've already applied.

werebus (approved):

PR #707, opened 2025-02-18
Add Current model to replace @current_user

werebus:


app/controllers/application_controller.rb
```diff
@@ -38,11 +36,13 @@ def redirect_unauthenticated
...
 
   def set_current_user
-    @current_user =
+    Current.user =
       if session.key? :user_id
         User.find_by id: session[:user_id]
       elsif session.key? :spire
-        User.find_by spire: session[:spire]
+        User.find_by(spire: session[:spire]).tap do |user|
+          session[:user_id] = user&.id
```
werebus, line 44: As far as I can tell, we've never set `session[:user_id]` before, so the `if` above was never used.

spec/views/layouts/application_spec.rb
```diff
@@ -3,19 +3,14 @@
 require 'rails_helper'
 
 describe 'layouts/application' do
-  context 'when current user is present' do
```
werebus, line 6: we removed the conditional logic from the navbar in #411. So folks see these links whether there's a "current user" or not.

dirschn (approved):

PR #841, opened 2025-09-16
Switch from yarn to npm

benmelz:
See https://github.com/umts/st-pax/pull/732


werebus:
> Furthermore this was taken care of by capistrano-yarn on screaming-dinosaur - have we just... not been bundling node modules on this app?

Yuuup.  `capistrano-yarn` was "unnecessary" in Rails 6.x, since Rails [did is for us](https://github.com/rails/rails/blob/86864c2f6661d3aa93b07f833b2900d241568cb7/railties/lib/rails/tasks/yarn.rake#L27-L30). So we removed it (#408) It should have been added back in the upgrade to Rails 7.0. We didn't notice because the Rails 7 upgrade was in January, 2024 (#552), but we haven't done anything except **remove** dependencies (#546) from `package.json` since October, 2023.

Seems like we probably should add JS to the Dependabot config on this repo, huh?

werebus (approved):

PR #867, opened 2025-09-24
Switch to cssbundling

benmelz:
Some odd notes:
- I went through the pain of getting the jquery-ui icon references to work in theory. I say in theory as I don't think we're using any of the ui widgets that depend on those icons in production? I might be wrong though and it's _relatively_ inexpensive to just set the paths up (and will show up in searches/will get thrown out when/if we drop it)
- We aren't importing bootstrap javascript at all? Which odd but makes my life easier.

### DIFF
--- a/spec/requests/users_spec.rb
+++ b/spec/requests/users_spec.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe 'UsersController' do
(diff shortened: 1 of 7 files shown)


werebus (changes requested):
That test failure seems like it would be my fault. I'm looking at it - it appears to be order-dependent. FWIW, #866 made that particular spec do anything at all; previously, there was an expectation with no `.to`

PR #1042, opened 2026-04-16
Lint javascript

werebus:
I was working on updating eslint, but @benmelz got in front of me (#1039). But I also had lint fixes in my branch, so here they are.

### DIFF
--- a/app/assets/javascripts/application_submissions/deny.js
+++ b/app/assets/javascripts/application_submissions/deny.js
@@ -1,6 +1,6 @@
 $(function() {
   $('form').on('change', '#application_submission_notify_of_denial', function(event) {
-    const inform = $(event.target).checked;
+    const inform = $(event.target).prop('checked');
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/application_submissions/deny.js
```diff
@@ -1,6 +1,6 @@
 $(function() {
-  $('form').on('change', '#application_submission_notify_of_denial', function(){
-    var inform = this.checked
+  $('form').on('change', '#application_submission_notify_of_denial', function(event) {
+    const inform = $(event.target).checked;
```
benmelz:
```suggestion
    const inform = $(event.target).prop('checked');
```

Haven't tested directly but I'm under the impression jquery doesn't implement `.checked` (previous `this` usage was actually using the native dom api).

benmelz (changes requested):

benmelz (approved):

PR #1135, opened 2026-07-23
Switch to solid_queue

KernSharma:
Replaces whenever with solid_queue. The daily move_to_dashboard task is now a job run from recurring.yml; solid_queue tables go in the main DB via migration. Workers run as a systemd service Capistrano restarts on deploy (service file added at merge). whenever fully removed.

Closes #1128


spec/jobs/move_to_dashboard_job_spec.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
```
benmelz: We can remove this, we're not really testing anything of substance here due to the mocks.

Gemfile
```diff
@@ -16,10 +16,10 @@ gem 'puma'
 gem 'rails', '~> 8.1.3'
 gem 'redcarpet'
 gem 'snappconfig'
+gem 'solid_queue'
```
benmelz: Should go in the production group

benmelz (changes requested):

spec/spec_helper.rb
```diff
@@ -6,6 +6,7 @@
 SimpleCov.start 'rails' do
   refuse_coverage_drop if ENV['CI']
   add_filter 'app/mailers/previews'
+  add_filter 'app/jobs'
```
benmelz: I don't think we should hardcode this. If we _need_ to get around the job being uncovered, inline comments in the job (i.e. `# simplecov:disable` ) (with the expectation that someday when we get full coverage the specs will be refactored) communicate a bit better that this is a bit of bandaid rather than how things should be

benmelz (approved):

PR #1136, opened 2026-07-24
Replace snappconfig with config_for and I18n

### LINKED ISSUE #617: Get rid of snappconfig
And our [`ApplicationConfiguration` wrapper](https://github.com/umts/jobapps/blob/f9605349cf45a4ee8df32886add3eb37a6669642/lib/application_configuration.rb). That gem hasn't been updated in 8 years (save for one README edit).

Values we have in the `application.yml` fall into four categories:

1. Actual application configuration, which we can replace with [Rails' built-in method](https://api.rubyonrails.org/classes/Rails/Application.html#method-i-config_for)
2. Things that should be an attribute on `Position`
3. A few site-wide strings, which could be moved into I18n
4. Things removed in #613

KernSharma:
Removes snappconfig and the ApplicationConfiguration wrapper. Application settings and behaviors are hardcoded (default_from in the mailer, site contact + organization name in I18n, the on_application_denial toggles made unconditional) rather than read from a config file, so the app no longer needs a hand-placed config file per deploy.

Closes #617


benmelz (changes requested):
I actually think that we don't really need a "config file" at all for these remaining values.

`on_application_denial` all dictates application behavior and realistically we only have this one deployed instance of `jobapps` - after 10 years or so of that being the case I think it's fine to just hardcode those behaviors.

Same for the contact emails - the default from is an `environment` setting in all of our other apps and site contact email could either be a translation or a direct hardcode wherever we use it (or in a helper).

benmelz (approved):

PR #1147, opened 2026-07-29
Add user entra_uid and backfill infrastructure

KernSharma:
First of the OAuth-migration PRs for #781, mirroring screaming-dinosaur #870. Adds the entra_uid column (nullable until the production backfill runs) with a unique index, the maintenance_tasks engine gated to admins, a Maintenance::ImportEntraUidsTask that backfills entra_uids from a spire_id→entra_uid CSV, and an admin-only spire-id CSV export (staff dashboard) to feed the PowerShell lookup.

Part of #781.


coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "line": 100.0
+    "line": 7.8
```
benmelz: Accidentally committed here

app/views/users/spire_ids.csv.erb
```diff
@@ -0,0 +1,5 @@
+<% require 'csv' %>
```
benmelz, line 1: This doesn't matter because this is just a temporary endpoint that'll be removed once we do the data processing but just to note

We typically prefer skipping view templating for csvs and instead prefer constructing them in the model layer directly using the ruby api.

But like I said we don't have to fix it for this one since it'll get removed hopefully soon

benmelz (changes requested):

spec/views/users/spire_ids_spec.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
```
benmelz: We shouldn't be writing view specs

It'd be appropriate to `# simplecov:disable` the csv export if needed since it's temporary

benmelz (changes requested):

benmelz (approved):

PR #1152, opened 2026-07-30
Add task to fill dummy entra uids

KernSharma:
Second of the OAuth-migration PRs for #781, mirroring screaming-dinosaur #885. FillMissingEntraUidsTask assigns placeholder entra_uids (missing-<spire>) to any users still nil after the real backfill — the inactive accounts the Azure lookup can't resolve. Stacked on #1147.

Part of #781.


benmelz (approved):
We can merge this into #1147 right now actually

PR #1155, opened 2026-08-03
Set up bootsnap

KernSharma:
Adds bootsnap to speed up boot by caching compiled Ruby and YAML. `gem 'bootsnap', require: false` plus `require "bootsnap/setup"` in boot.rb, matching the standard Rails/screaming-dinosaur setup.

Closes #1145.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@
 source 'https://rubygems.org'
 ruby file: '.ruby-version'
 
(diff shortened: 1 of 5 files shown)


benmelz (changes requested):
One more thing - we need to add a precompile step to the deploy process for the current production environment (i.e. a capistrano task)

benmelz (approved):

PR #1168, opened 2026-08-12
Switch to OAuth

### LINKED ISSUE #781: Switch from SAML to OAuth

KernSharma:
Switches jobapps from Shibboleth to OAuth (Microsoft Entra ID), mirroring
screaming-dinosaur #881. Last of the #781 stack.

entra_uid is now the required identity. Login runs through the OmniAuth
callback and logout through Entra; auth is keyed on the session entra_uid
instead of Shibboleth headers.

spire stays but is now optional; it gets dropped, with the promote rework,
in a follow-up that mirrors SD #886. The backfill scaffolding from the
earlier stack PRs is removed here since entra_uid is now required.

Closes #781


benmelz (changes requested):
Note that I pushed the necessary credentials directly here so make sure to `git pull` before continuing locally

---

This is a good start, however I would like to authentication/access control flow ironed out as a part of this PR as well.

_Currently_ we are using a dedicated endpoint that we redirect to for the login forms (`/sessions/unauthenticated`) which isn't so ideal due to the fact that this obfuscates what page the user initially visited and they won't be sent back there after authenticating with MS.

The model that we've been switching to is as follows:
1. Request comes in to any endpoint
2. Look up the user
3. Run access control logic for the current page, and _raise an error_ if denied for any reason (specific to the endpoint)
4. Rescue from that error globally
  a. If they don't have a user/session record, render the login page (as a response to the initial request with a 401 status)
  b. If they do have a user/session record, render a forbidden page (as a response to the initial request with a 403 status)

I would take cues from [screaming-dinosaur's `Authorizable` concern](https://github.com/umts/screaming-dinosaur/blob/main/app/controllers/concerns/authorizable.rb) which implements this flow pretty nicely.

Some caveats:
- We already have ActionPolicy set up there which inherently provides 3 and 4 (the `authorize!` system and the `ActionPolicy::Unauthorized` error). You'll have to implement an alternative here:
  - just add a thin error class at `app/errors/unauthorized.rb` to use
  - raise that error in your analog to `ApplicationController#access_control` instead of calling `deny_access`, as well as any other more restrictive access control callbacks in other controllers (search for other usages of `deny_access`)
- We have a dedicated user registration page there that people get redirected to. The model is a bit different here (you are allowed in as long as you have a session and your user record isn't created until you submit your application). Just keep things how they are for now (we'll probably tackle this in a followup PR).

app/views/application/production_login.haml
```diff
@@ -0,0 +1,2 @@
+%h1 Log in
```
benmelz, line 1: This should be a blank page that has an invisible automatically-submitted form on it

See https://github.com/umts/round-three/blob/master/app/views/shared/production_login.html.haml

benmelz (changes requested):

